PR #26, opened 2018-02-12
Rework of the German translation

MarwinPhilips:
What's written in the title is the acutall content, changing most of the german textfiles.


badgamernl:
Thanks!

PR #29, opened 2018-03-20
Corrected afk time in kick reason

Jon8RFC:
I recently saw a user (under one hour), kicked for being afk for 36,000 minutes.

afk_kick


Cooldude2606:
This change has been resolved already, but it is nice that you attempted to fix it your self before we noticed.

PR #30, opened 2018-04-14
Swedish translation (reports)

sakamakj:
No description needed


sakamakj:
Some testing, thats why there is upload and deleting

Cooldude2606:
Great to see more translations being added!

PR #31, opened 2018-04-14
Lots of corrections from Jon8RFC

Cooldude2606:


Cooldude2606:
Nice to see some changes being made even if no pull request was maed by Jon8RFC.

PR #34, opened 2018-04-18
Delete reports.cfg

sakamakj:


sakamakj:
file with wrong path deleted

PR #35, opened 2018-04-18
Fixed most unlikely edgecase

Valansch:
In case of numerical username


Cooldude2606:
Very Nice :D But as i said on the discord is this really ever going to happen... still great that you spent a good 5 minutes making this :P

Valansch:
I wouldn't have been mad if you rejected it. I don't even know if its worth the extra complexity :D

PR #36, opened 2018-04-19
translations fixed (all files but one I think)

sakamakj:
Read comments, I also added the file that isnt translated into sv-SE for the sake of completion. Can have missed something.


Cooldude2606:
many Translations; Much Wow

PR #42, opened 2018-05-10
Fix repair command

trumank:
I have...
- optimized /repair by making only one call to find entities and filtering out entities outside of the repair radius
- fixed repairing entities not sharing a force with the player which prevents healing enemies or trees/rocks
- fixed reviving ghosts which were placed by the player (not from being destroyed) which allowed for repairing technologies that were not researched or were otherwise unavailable


Cooldude2606:
Great less work for me :D

PR #43, opened 2018-05-10
Fix body marker not appearing when bonus respawn is applied

trumank:


Cooldude2606:
Even more work i dont need to do :D

Issue #58, opened 2018-05-23
Add /tp command back


XenoCyber88:
The new teleport commands are overly complicated and require you to remember two separate commands. 

please re-add the /tp command for power users mods so they can quickly simply teleport to other users 

/tp from to

Cooldude2606:
"overly complicated" /go-tp \<name\> or /bring \<name\> 
![NOTP](https://cdn.discordapp.com/emojis/447484775616348170.png?v=1)

sakamakj:
I agree with Cooldude2606. /go-to player and /bring player works really well and if you change that it will take longer time for me instead.  It is really fast when using tab-completion and nowadays you can also use the /set-home command before you use it so that you can return after investigating the potential trouble.

wsfjm:
I think too many vocabularies do not have localized files.
I'm sorry I can't read English at all. Only with translation software!

Cooldude2606:
> do not have localized files

We understand that problem but out side the ones already done we have not had any one willing to add more to it. wsfjm if you wish to add your first language to it then it is fine by me.
See: https://help.github.com/articles/creating-a-pull-request/

wsfjm:
I will not, Creating a pull request，https://github.com/wsfjm/explosivegaming-main/tree/master/locale/zh-CN[](url)
Not enough vocabulary

JCA122204:
Mabye bring /tp back and leave the others, so that all 3 are available?

XenoCyber88:
That had always been my request not to move one for the other but have all
coexist

Cooldude2606:
Please see: #67

PR #65, opened 2018-05-31
Add /tp Command To Co-Exist With /bring And /go-to

JCA122204:
#58 

And,if there is something wrong,please tell me. I am trying my best to learn this so that I can be of more help in the future


JCA122204:
/go-to And /bring Don’t Include Any Of Those Checks Either

PR #72, opened 2018-06-09
Fix logic error on table sort function

flbuddymooreiv:
This is used probably elsewhere but the warp list is the main place this error manifests whereby strings are sorted by ascii code rather than lexicographically. This is not wrong per se, but I do not think it was the intended behavior and the scenario gameplay is a bit impacted by the current sort order. This PR resolves the issue.

### DIFF
--- a/StdLib/table.lua
+++ b/StdLib/table.lua
@@ -246,18 +246,20 @@ function table.deepcopy(object)
     return _copy(object)
 end
 
+--- Default table comparator sort function.
+-- @local
+-- @param x one comparator operand
+-- @param y the other comparator operand
+-- @return true if x logically comes before y in a list, false otherwise
 local sortfunc =
     function(x, y) --sorts tables with mixed index types.
         local tx = type(x)
         local ty = type(y)
         if tx == ty then
-            if type(x) == 'string' and type(y) == 'string' then
-                if string.lower(x) == string.lower(y) then
-                    return x < y
-                end
+            if type(x) == 'string' then
                 return string.lower(x) < string.lower(y)
             else
-                return x < y and true or false --similar type can be compared
+                return x < y
             end
         elseif tx == 'number' then
             return true --only x is a number and goes first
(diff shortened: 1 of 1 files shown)


Cooldude2606:
There is a better function already made called table.alpanumsort see StdLib/table.lua ln 503; however we thought that this may be too heavy of a process to use as a result of the heavy use of string manipluators. 
Would you consider looking over this function and if this one matches the same outcome that you are hoping for could you use the one there, also if this one does fit the same need as the one you made i am sorry for the work you put into it, if yours still has an advantage over this one then ill merge the branch.
I will wait for a reply back from you.

flbuddymooreiv:
Thank you for your consideration. The effort was more in the workflow and
lua bootstrap than anything else. I'd been meaning to play with lua anyway,
so no worries there. The function is not a new function but rather a
deduplicated and slightly modified existing in line function, so no worries
there either. In reality, it looks like this function that appears to be
new is actually being called every time already, so I would not expect a
performance issue. However, I do notice that I injected a bug with respect
to the number type check so perhaps I should try a new PR? Will you still
consider the minor refactor?

Cooldude2606:
It is up to you if you want to make a new PR and close this one, i have not looked into detail at the changes you are made, only a gernal overview. I will look in detail at the changes you have made once i get the time and will look out for this bug you metioned which was introduced. 
Once i look over the code i may then merge the branch, howevver it is compelty up to you if you close it before then (for example if you would like to make drastic changes) and later reopen it.

flbuddymooreiv:
I resolved the issue with the other error. I didn't realize on PR could have additional changes after submission. Does this look comprehensible on your end?

flbuddymooreiv:
I've incorporated your changes.. Unfortunately, pulling the duplicated code into a function probably made it look like I authored more than I did, but I think the things you pointed out were worth fixing, so hopefully they qualify as acceptable now.

Issue #111, opened 2019-07-02
2 chat commands not working


CmonMate497:
**Factorio version**
Version number: 0.17.53

**Scenario version**
Version number: V5.8

**Bug description**
rhd/lhd and redmew chat command responds with Unknown key

**Reproduction method**
1. Launch Factorio
2. Go onto an ExpGaming server or launch scenario in single player
3. Type in rhd, lhd and redmew
4. See error in log

**Expected behavior**
To respond with something like for lhd, "Trains Must be in LHD"

**Additional context**
Probably not high priority but thought I'd point it out

Cooldude2606:
See pull request #112

PR #112, opened 2019-07-02
Update addons.cfg

CmonMate497:
Added lhd and redmew chat bot so it won't appear with Unknown key error. Feel free to edit the messages as this was a quick change ;)


Cooldude2606:
See issue #111

locale/en/addons.cfg
```diff
@@ -56,4 +56,6 @@ aids=≖ ‿ ≖ Fear the aids of a public server ≖ ‿ ≖
 riot=(admins) ┬┴┬┴┤ᵒ_ᵒ)├┬┴┬┴ ‹ ‹\(´ω` )/››‹‹\ ( 　´)/››‹‹\ ( ´ω`)/›› (rest of server)
 loops=NO LOOPS; LOOPS ARE BAD; JUST NO LOOPS!!!!!; IF YOU MAKE A LOOP.... IT WILL NOT END WELL!!!!!!!
 lenny=( ͡° ͜ʖ ͡°)
-hodor=Hodor
\ No newline at end of file
+hodor=Hodor
+lhd=Trains MUST be in LHD.
```
Cooldude2606: Would it not have been better to edit the config file to use info.lhd and info.redmew rather than adding new keys that would require further translating?

locale/en/addons.cfg
```diff
@@ -56,4 +56,6 @@ aids=≖ ‿ ≖ Fear the aids of a public server ≖ ‿ ≖
...
+redmew=We don't talk about Redmew here, they beat us to 1000 members =:(
```
Cooldude2606: Would it not have been better to edit the config file to use info.lhd and info.redmew rather than adding new keys that would require further translating?

CmonMate497: rip I didn't see that it was already in info xD

PR #113, opened 2019-07-03
Spelling fixes

bbassie:
Warp list has some spelling errors and the copying of the task list has left some unnamed comments and locale keys.


bbassie:
Most of the spelling fixes that where underlined by cspell

bbassie:
Still needs to be tested if it fully works but don't expect it to have major issues

PR #122, opened 2019-08-09
Changed my color

zaanposni:
As in #120


zaanposni:
Changed ArPiiXs color aswell as he requested it

PR #128, opened 2019-08-23
Spelling errors in this folder

CmonMate497:
Didn't fix script-wrning-removed as I'm blind and can't find where that's needed in the script. Everything else should be fine


CmonMate497:
if you can find the thing that I'm too blind to see, then go ahead and add it in if you wish. doesn't make a difference though 🤷‍♂

PR #135, opened 2020-01-28
Ratio command(on dev now)

tovernaar123:
Added a ratio command to calculate the input and ouput ratios for the selected machine supports, moduals as well. (on dev now)


modules/commands/ratio.lua
```diff
@@ -0,0 +1,85 @@
+
+
+local Commands = require 'expcore.commands'
+
```
Cooldude2606: Could have used a require for general parse but this in non blocking.

Issue #80, opened 2018-09-03
Auto Station Names


Cooldude2606:
**Is your feature request related to a problem? Please describe.**
Would be a nice idea to have stations automatically name them selfs based on the ores around them or proximity to certain entities like spawn or a semting array.

**Describe the solution you'd like**
This would enforce a set way of naimng stations being: [U/L/#-Ore/Pax] [Dir] \<baker name\>
Few examples:
[U-Fe] {NW-1} arty714
[L-Cu] {S-2} cooldude2606
[#-PAX] {W-1} mark9064

Cooldude2606:
Added in PR #137

PR #137, opened 2020-02-10
adding auto-station naming (suggested by cooldude) try 2

tovernaar123:
also the ratio command needs some lines in the cfg file which i messed up last time


modules/addons/station-auto-name.lua
```diff
@@ -0,0 +1,70 @@
...
+            --Check which recource is closest
+            for i, item in ipairs(recoursec) do
+                if not closest_distance then
+                    recourse_closed = item
+                    local dx, dy = px - item.bounding_box.left_top.x, py - item.bounding_box.left_top.y
+                    closest_distance = (dx*dx)+(dy*dy)
+                else 
+                    local dx, dy = px - item.bounding_box.left_top.x, py - item.bounding_box.left_top.y
+                    if   (dx*dx)+(dy*dy) < closest_distance then
+                        closest_distance = (dx*dx)+(dy*dy)
```
Cooldude2606: To save compute time please use a varible for `(dx*dx)+(dy*dy)`
You also this block of code is why similar to the block above it, try: `if not closest_distance  or  (dx*dx)+(dy*dy) < closest_distance then`

modules/addons/station-auto-name.lua
```diff
@@ -0,0 +1,70 @@
...
+                    end
+                end
+        
+            end
+            
+
+            local item_name = recourse_closed.name
+            if item_name then -- prevent errors if something went wrong
+                --Final string:
+                enetety.backer_name = string.format("[L] [img=item.%s] %s %s (%s)",item_name,item_name:gsub("^%l", string.upper):gsub('-',' '),enetety.backer_name,Angle( enetety ))
```
Cooldude2606: Nothing wrong with this but i would suggest taking `item_name:gsub("^%l", string.upper):gsub('-',' ')` out as a varible in the line before to make it more human readible.

modules/addons/station-auto-name.lua
```diff
@@ -0,0 +1,70 @@
...
+end
+--add func to robot and player build entities
+Event.add(defines.events.on_built_entity,station_name_changer)
+Event.add(defines.events.on_robot_built_entity,station_name_changer)
+
+    
+--Credit to Cooldude2606 for using his lua magic to make this function.
+function Angle( enetety )
+    local angle = math.atan2(enetety.position.y,enetety.position.x)/math.pi
+    local directions = {
```
Cooldude2606: I would move this table declaration outside of the function to avoid needing to re-declare it every time the function is used.

Cooldude2606:
This PR is for feature request: Auto Station Names #80

modules/addons/station-auto-name.lua
tovernaar123: lol ok

tovernaar123:
i fixed it

PR #143, opened 2020-02-27
feature/tree-decon

Cooldude2606:
Added fast tree removeal which will remove trees when deconed; can be used with donator perks. Has minimal lag when removing.

### DIFF
--- a/modules/addons/tree-decon.lua
+++ b/modules/addons/tree-decon.lua
@@ -1,0 +1,61 @@
+--- Makes trees which are marked for decon "decay" quickly to allow faster building
+-- @addon Tree-Decon
+
+local Event = require 'utils.event' --- @dep utils.event
+local Game = require 'utils.game' --- @dep utils.game
+local Global = require 'utils.global' --- @dep utils.global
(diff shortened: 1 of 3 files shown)


modules/addons/tree-decon.lua
```diff
@@ -0,0 +1,61 @@
...
+    chache[player.name] = true
+
+    -- Check that the entity is allowed to be isntant deconed
+    local head = tree_queue._head + 1
+    local entity = event.entity
+    if entity and entity.valid and not entity.last_user and entity.type ~= 'cliff' then
+        tree_queue[head] = entity
+        tree_queue._head = head
+    end
+end)
```
bbassie: Is it not better to use event filter to filter out the unneeded entities rather than checking all of them?
https://lua-api.factorio.com/latest/Event-Filters.html#LuaEntityMarkedForDeconstructionEventFilters

Cooldude2606: Event filters can not be used in softmod scenarios because of being limited to only one event handler; this one event handler is used by Event to allow multiple handlers to be registered. So sadly we still need to check using api calls.

PR #156, opened 2020-05-16
Feature/loadquickbar

kevintayloruk:
Added a /load-quickbar command:

This reads a pre-set list from config/preset_player_quickbar.lua. {1..100} where 1-10 is quickbar 1, 11-20 is quickbar 2, etc. These use the factorio item names.

Added command file to the loader.
Added command to the sponsor role.

### DIFF
--- a/modules/commands/quickbar.lua
+++ b/modules/commands/quickbar.lua
@@ -35,7 +35,7 @@ Commands.new_command('save-quickbar','Saves your Quickbar preset items to file')
     for i=1, 100 do
         local slot = player.get_quick_bar_slot(i)
         if slot ~= nil then
-            table.insert(quickbar_names, i, slot.name)
+            quickbar_names[i] = slot.name
         end
     end
     game.write_file("quickbar_preset.txt", player.name .. " = " .. serpent.line(quickbar_names) .. "\n", true)
--- a/modules/commands/quickbar.lua
+++ b/modules/commands/quickbar.lua
@@ -1,0 +1,26 @@
+--[[-- Commands Module - Quickbar
+    - Adds a command that allows players to load Quickbar presets
+    @commands LoadQuickbar
+]]
+
+local Commands = require 'expcore.commands' --- @dep expcore.commands
+local Roles = require 'expcore.roles' --- @dep expcore.roles
+local Game = require 'utils.game' --- @dep utils.game
+local config = require 'config.preset_player_quickbar' --- @dep config.preset_player_quickbar
+require 'config.expcore.command_general_parse'
+
+
+--- Changes the amount of bonus you receive
+-- @command bonus
+-- @tparam number amount range 0-50 the percent increase for your bonus
+Commands.new_command('load-quickbar','Loads your preset Quickbar items')
+:register(function(player)
+    if config.players[player.name] then
+        local custom_quickbar = config.players[player.name]
+        for i, item_name in ipairs(custom_quickbar) do
+          player.set_quick_bar_slot(i, item_name)
+        end
+    else
+        Commands.print('Quickbar preset not found','red')
+    end
+end)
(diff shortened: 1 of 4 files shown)


kevintayloruk:
Think I've addressed both of those now @tovernaar123

modules/commands/quickbar.lua
```diff
@@ -0,0 +1,25 @@
+--[[-- Commands Module - Quickbar
+    - Adds a command that allows players to load Quickbar presets
+    @commands LoadQuickbar
+]]
+
+local Commands = require 'expcore.commands' --- @dep expcore.commands
+local Roles = require 'expcore.roles' --- @dep expcore.roles
+local Game = require 'utils.game' --- @dep utils.game
+local config = require 'config.preset_player_quickbar' --- @dep config.preset_player_quickbar
+require 'config.expcore.command_general_parse'
```
Cooldude2606, line 10: This config isnt required for your command to work. This config adds stuff like number and player parse for command parameters. Since you dont have any parameters you dont need this config.

modules/commands/quickbar.lua
```diff
@@ -0,0 +1,25 @@
...
+
+
+--- Loads your quickbar preset
+-- @command load-quickbar
+Commands.new_command('load-quickbar','Loads your preset Quickbar items')
+:register(function(player)
+    if config[player.name] then
+        local custom_quickbar = config[player.name]
+        for i, item_name in ipairs(custom_quickbar) do
```
Cooldude2606, line 19: I would always suggest using ipairs as it is faster; however, in this case you might have slots 1-5 with filters and then slots 11-15 with filters. By using ipairs the loop would stop once it reaches 5 as there is no index 6. This then becomes an issue and you should use pairs.

kevintayloruk: Using pairs and checking for nil / '' value

modules/commands/quickbar.lua
kevintayloruk: Removed the require.

kevintayloruk:
Resolved 2 changes that were requested.

Never I've added a /save-quickbar command and this writes to a local script-output/quickbar_preset.txt file.

kevintayloruk:
Added /load-toolbar and /save-toolbar aliases.
Saves output as a table to file.

modules/commands/quickbar.lua
```diff
@@ -0,0 +1,44 @@
+--[[-- Commands Module - Quickbar
+    - Adds a command that allows players to load Quickbar presets
+    @commands LoadQuickbar
+    @commands SaveQuickbar
```
Cooldude2606, line 4: The tag `@commands` can only be used once. It is used to show this module is a command module rather than a gui or control module. I would suggest `@commands Quickbar`

modules/commands/quickbar.lua
```diff
@@ -0,0 +1,44 @@
...
+--- Saves your quickbar preset to the script-output folder
+-- @command save-quickbar
+Commands.new_command('save-quickbar','Saves your Quickbar preset items to file')
+:add_alias('save-toolbar')
+:register(function(player)
+    local quickbar_names = {}
+    for i=1, 100 do
+        local slot = player.get_quick_bar_slot(i)
+        if slot ~= nil then
+            table.insert(quickbar_names, i, slot.name)
```
Cooldude2606, line 39: This is a small performance thing. table.insert is actully quite slow and should only be used when inserting into the middle of an array like table. Since you already have a for loop, meaning ever index is greater than the last, you can use `quickbar_names[i] = slot.name`.
For some extra info on performance: if you need to insert at the end of an array then the fastest option is `tbl[#tbl+1] = value` since `#tbl` is the length of the array and `#tbl+1` is index which is not yet used.

modules/commands/quickbar.lua
kevintayloruk: Fixed

modules/commands/quickbar.lua
kevintayloruk: I'd actually changed this locally for clarity but wasn't aware of the performance issue. Fixed.

kevintayloruk:
Pulled the vastest luacheckrc.

PR #157, opened 2020-05-21
Move banned player inv to spawn chests

### LINKED ISSUE #155: clear inv on ban
**Feature overview**
Name: name-here
TL;DR: a few words about what the feature is
clear inventory of player on when banned
**Details**
A detailed description of what the new feature is.
nothing to explain XD
**Possible methods**
Any ideas you have had about how to implement the feature or any problems that may arise.

**Additional context**

kevintayloruk:
Altered ban callback to include the call to the move_items function as per temp_ban.
This fixes #155


tovernaar123:
this solves my isue thnx 👍

Cooldude2606:
One issue with this is that using /ban wont clear the players inventory. There is the on_player_banned event which can be used but im not sure where this would fit. Another possible fix could be to have the clear inventory command work on offline players, but you will still need to check the player is alive.

Cooldude2606:
After further thought I would suggest a new module which uses on_player_banned and on_player_kicked to move the inventory rather than just through the gui methods.

kevintayloruk:
Agreed. Closing this pr.

PR #160, opened 2020-05-26
Added Auto Doc Generation and PR Checking

Cooldude2606:
Added github action that will generate documentaion following the ldoc style automaticly when anything is push to the dev branch.
Added github action that will check prs with luacheck to look for errors and also checks that there are no doc comment errors.
Removed factorio luacheckrc as a submodule in favour of having a custom luacheckra and pr checking which doesnt fail every time.
Fixed all existing errors that were found by luacheck.


Cooldude2606:
Re: luacheck fail
The luacheck failed on this pr as the luacheckrc file has not yet been updated which is part of the changes in this pr.

PR #173, opened 2020-08-16
Added ability to warp while driving in a vehicle

bbassie:
This adds the ability to stay in your vehicle while warping.

Wanted to do this for ease of use with the Spidertron but it is not able to be teleported yet, I've submitted a bug report:
[bug report](https://forums.factorio.com/viewtopic.php?f=7&t=88128)

### DIFF
--- a/modules/control/warps.lua
+++ b/modules/control/warps.lua
@@ -273,10 +273,28 @@ function Warps.teleport_player(warp_id, player)
         y=warp.position.y+0.5
     }
 
-    -- Teleport the player
-    local goto_position = surface.find_non_colliding_position('character', position, 32, 1)
-    if player.driving then player.driving = false end
-    player.teleport(goto_position, surface)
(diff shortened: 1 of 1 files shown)


bbassie:
Spider feature comming soon: (reply from: Rseding91)
It was marked as not teleportable shortly before the release due to complications with how it worked; it didn't when teleported. So, it's intentional.

However; two days ago I implemented teleporting for them so it will be in some future release.

Cooldude2606:
Has this been made possible yet?

bbassie:
It should be working, if I'm not mistaking it worked last time I tried to do it after the update where they added it.

Cooldude2606 (approved):
All your changes are good and work correctly. The force pushes were used to remove the merges I preformed before you made your changes. These merges from the release branch only created merge confilcts and should not have been done.

PR #174, opened 2020-08-17
Delete outdated feedback link

bbassie:
- Doesn't seem to be used anymore and obstructs the view when at spawn.


Cooldude2606 (approved):
Looks good to me, only 3 line removals.

PR #175, opened 2020-08-17
Added Autofill toolbar button

bbassie:
Add Autofill ability draft, going to add more commits to finalize to a final product.


config/gui/autofill.lua
```diff
@@ -0,0 +1,93 @@
...
+			enabled = false
+		},
+		{
+			type = 'fuel',
+			inventories = {defines.inventory.fuel},
+			item = 'coal',
+			amount = 10,
+			enabled = false
+		}
+	}
```
bbassie, line 92: Not sure about this config layout may still need to be changed

modules/gui/autofill.lua
```diff
@@ -0,0 +1,192 @@
...
+                    inserted = entity_inventory.insert({name=item, count=item_amount})
+                    player_inventory.remove({name=item, count=inserted})
+                    print_text(entity.surface, entity.position, {'autofill.filled', '[img=entity/'..entity.name..']', inserted, '[img=item/'..item..']' }, { r = 255, g = 165, b = 0, a = 1})
+                end
+                goto continue
+            end
+            ::continue::
+        end
+    end
+end
```
bbassie, line 190: This part is also going to need some changes because the performance is not optimal also because of the config.

modules/gui/autofill.lua
```diff
@@ -0,0 +1,192 @@
...
+end)
+
+local function inventories_contains_inventory(inventories, inventory)
+    for _, inv in pairs(inventories) do
+        if inv == inventory then
+            return true
+        end
+    end
+    return false
+end
```
bbassie, line 142: Temporary function should't need it when config changed.

config/gui/autofill.lua
```diff
@@ -0,0 +1,93 @@
...
+				enabled = true
+			}
+		},
+		['locomotive'] = {
+			{
+				type = 'fuel',
+				inventory = defines.inventory.fuel,
+				enabled = true
+			}
+		},
```
bbassie, line 26: Locomotive part not working need to figure out why.

bbassie (commented):
Comments on the current status of the PR

Cooldude2606 (changes requested):
General note about some functions: where you have function of the form `print({...})` you can change it to `print{...}` other examples include `can_insert` and `insert`

config/gui/autofill.lua
```diff
@@ -0,0 +1,216 @@
...
+local shell = 'shell'
+
+local car = 'car'
+local tank = 'tank'
+local spidertron = 'spidertron'
+local locomotive = 'locomotive'
+local gun_turret = 'gun-turret'
+local burner_mining_drill = 'burner-mining-drill'
+local stone_furnace = 'stone-furnace'
+local steel_furnace = 'steel-furnace'
```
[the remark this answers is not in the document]
bbassie: ecadbd6

modules/gui/autofill.lua
```diff
@@ -0,0 +1,362 @@
...
+--- Used to assign an event to the header label to trigger a toggle
+-- @element header_toggle
+local header_toggle = Gui.element()
+:on_click(function(_, element, event)
+    event.element = element.parent.alignment[toggle_section.name]
+    toggle_section:raise_custom_event(event)
+end)
+
+--- Used to assign an event to the header label to trigger a toggle
+-- @element header_toggle
```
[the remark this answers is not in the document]
bbassie: e23092a

modules/gui/autofill.lua
```diff
@@ -0,0 +1,362 @@
...
+-- All entity autofill settings will be ignored if its disabled
+-- @element toggle_item_button
+local toggle_section =
+Gui.element{
+	type = 'sprite-button',
+	sprite = 'utility/expand_dark',
+	hovered_sprite = 'utility/expand',
+	tooltip = {'autofill.toggle-section-tooltip'}
+}
+toggle_section
```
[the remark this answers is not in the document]
bbassie: 620051d

modules/gui/autofill.lua
```diff
@@ -0,0 +1,362 @@
...
+local Event = require 'utils.event' --- @dep utils.event
+local table = require 'overrides.table' --- @dep overrides.table
+
+local print_text = Game.print_floating_text -- (surface, position, text, color)
+
+--- Table that stores if autofill is enabled or not
+local autofill_player_settings = {}
+Global.register(autofill_player_settings, function(tbl)
+    autofill_player_settings = tbl
+end)
```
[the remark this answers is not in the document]
bbassie: Maybe later but is hard to test locally for me so you can add it if its needed.

config/gui/autofill.lua
```diff
@@ -0,0 +1,216 @@
...
+		locomotive = locomotive,
+		gun_turret = gun_turret,
+		burner_mining_drill = burner_mining_drill,
+		stone_furnace = stone_furnace,
+		steel_furnace = steel_furnace
+	},
+	default_entities = {}
+}
+
+local default_autofill_item_settings = {
```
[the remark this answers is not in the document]
bbassie: f149e44

locale/en/gui.cfg
```diff
@@ -81,6 +81,17 @@ edit-tooltip=Currently being edited by: __1__
...
+[autofill]
+main-tooltip=Autofill settings
+toggle-section-caption=__1__ __2__
+toggle-section-tooltip=Expand Section
+toggle-section-collapse-tooltip=Collapse Section
+toggle-entity-tooltip=Toggle the autofill of __1__ entity (overwrites autofill settings of the entity)
+toggle-tooltip=Toggle the autofill of __1__ into the __2__ slot of entities
+amount-tooltip=Amount of items it will try and put into the __1__ slot of __2__
+invalid=Set autofill amount to the maximum __1__ __2__ for __3__
+filled=Autofilled the __1__ with __2__ __3__
```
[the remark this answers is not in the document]
bbassie: e52f06d

config/gui/autofill.lua
```diff
@@ -0,0 +1,216 @@
...
+generate_default_setting(config.entities.gun_turret, defines.inventory.turret_ammo, true)
+
+generate_default_setting(config.entities.burner_mining_drill, defines.inventory.fuel, true)
+
+generate_default_setting(config.entities.stone_furnace, defines.inventory.fuel, true)
+
+generate_default_setting(config.entities.steel_furnace, defines.inventory.fuel, true)
+
+-- Cleanup temporary table
+default_autofill_item_settings = nil
```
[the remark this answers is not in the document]
bbassie: Removed it altogether now.

modules/gui/autofill.lua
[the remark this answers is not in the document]
Cooldude2606: Testing locally is not effected by the data sync. Because you are not connected to the server a fake responce will be triggered internally after 5 seconds to ensure that scripts still work. But I agree it can be done later as data syncing is not a requirement.

modules/gui/autofill.lua
```diff
@@ -0,0 +1,361 @@
...
+Gui.element(function(_, parent, i)
+    local toggle_flow = parent.add{ type = 'flow', name = 'toggle-setting-empty-'..i }
+    local amount_flow = parent.add{ type = 'flow', name = 'amount-setting-empty-'..i }
+    toggle_flow.style.padding = 0
+    amount_flow.style.padding = 0
+    local toggle_element = toggle_item_button_empty(toggle_flow, i)
+    toggle_element.enabled = false
+    local amount_element = amount_textfield_empty(amount_flow, i)
+    amount_element.enabled = false
+end)
```
Cooldude2606, line 236: Do toggle_item_button_empty and amount_textfield_empty require names? Do they even require the flows?
I suggest removing the flows and removing the names from the empty elements so factorio will auto assign a unique name. This will also mean that you dont need to use a function to define the empty elements.

Cooldude2606 (changes requested):
Everything else is good, just one change suggested.

modules/gui/autofill.lua
bbassie: ef823f2

Cooldude2606 (changes requested):
I thought I had already pointed this out but I guess I missed it. You use string.len on string literals which can be simplified to constants. These apear on lines 75, 133, and 175. Everything else is fine.

locale/en/gui.cfg
```diff
@@ -81,6 +81,17 @@ edit-tooltip=Currently being edited by: __1__
 edit-tooltip-none=Currently being edited by: Nobody
 discard-tooltip=Remove task
 
+[autofill]
+main-tooltip=Autofill settings
+toggle-section-caption=__1__ __2__
+toggle-section-tooltip=Expand Section
+toggle-section-collapse-tooltip=Collapse Section
+toggle-entity-tooltip=Toggle the autofill of __1__s
```
Cooldude2606, line 89: My previous suggestion of adding the s on the end was wrong. I did not know that you were using rich text for `__1__` which leeds to a strange looking tooltip.
tooltip

modules/gui/autofill.lua
```diff
@@ -0,0 +1,348 @@
...
+    scroll_table.parent.style.padding = 0
+    -- Remove the default gap that is added in a table between elements
+    scroll_table.style.vertical_spacing = 0
+    -- Loop over each default entity config
+    for _, setting in pairs(config.default_entities) do
+        local table_sizes = {}
+        local tables = {}
+        -- Draw a section for the element
+        local entity_table = section(scroll_table, setting.entity, 3)
+        entity_table.style.padding = 3
```
Cooldude2606, line 242: I would suggest aligning the table to be centered to avoid a large gap on one side.
gap

modules/gui/autofill.lua
```diff
@@ -0,0 +1,348 @@
...
+    if not setting then return end
+    if setting.enabled then
+        setting.enabled = false
+        element.sprite = 'utility/close_black'
+        element.style = 'shortcut_bar_button_red'
+    else
+        setting.enabled = true
+        element.sprite = 'utility/confirm_slot'
+        element.style = 'shortcut_bar_button_green'
+    end
```
Cooldude2606, line 87: I think that setting the style here is resetting the padding causing the icons for the cross and tick to be different sizes. If not then it might be best to having a different padding for each icon to make them apear the same size.
sizes

modules/gui/autofill.lua
```diff
@@ -0,0 +1,348 @@
...
+    -- Draw the internal container
+    local container = Gui.container(parent, event_trigger)
+    -- Draw the scroll container
+    local scroll_table = Gui.scroll_table(container, 400, 1, 'autofill-scroll-table')
+    -- Set the scroll panel to always show the scrollbar (not doing this will result in a changing gui size)
+    scroll_table.parent.vertical_scroll_policy = 'always'
+    -- Scroll panel has by default padding
+    scroll_table.parent.style.padding = 0
+    -- Remove the default gap that is added in a table between elements
+    scroll_table.style.vertical_spacing = 0
```
Cooldude2606, line 235: I think there may still be some horizontal spacing which is causing these gaps on both sides. These gaps show both when the sections are closed and when they are open.
gaps

Cooldude2606 (changes requested):
All of the code is now fine, there are just some style issues that need addressing.

modules/gui/autofill.lua
bbassie: I think you've got an older version because fixed this some commits ago. And cannot reproduce.

modules/gui/autofill.lua
bbassie: They have the same size and padding is pixel perfect the same for both so I don't see any change needed.
ShareX_tshXrSEgRi
(padding top and bottom is also the same for both buttons!)

modules/gui/autofill.lua
bbassie: They are centered now:
factorio_OpiKeTuIuK

locale/en/gui.cfg
bbassie: https://github.com/explosivegaming/scenario/pull/175/commits/afcd16c56e30fcb3bc3bd3bf7935c4fc9b952ce4

Cooldude2606 (changes requested):
Autofill is not working, the toggle_item_button does not change state when clicked.

Cooldude2606:
Last suggestion and then I will be happy to merge: increase the space between the inserted messages so that it is closer to what the base game pickup messages are. (images are at different zoom levels)
zoomOne zoomTwo

bbassie:
> Last suggestion and then I will be happy to merge: increase the space between the inserted messages so that it is closer to what the base game pickup messages are. (images are at different zoom levels)
> zoomOne zoomTwo

Fixed: https://github.com/explosivegaming/scenario/pull/175/commits/c46244e7edf7adc3df449c12077380ce21405960
image

Cooldude2606 (approved):
Everything looks good now.

PR #177, opened 2020-09-06
Painting with colored concrete

bbassie:
Adds a GUI for painting with the new colored concrete:
Paint preview

This is a basic implementation with some dirty workarounds:
- Uses deactivated flying text entities to store the tiles that were below...
    - Thought about doing it in global but then I think it would just low the loading and saving too much (maybe worth trying)

### DIFF
--- a/config/_file_loader.lua
+++ b/config/_file_loader.lua
@@ -59,6 +59,7 @@ return {
     'modules.gui.warp-list',
     'modules.gui.task-list',
     'modules.gui.player-list',
+    'modules.gui.paint',
     'modules.gui.server-ups',
     'modules.commands.debug',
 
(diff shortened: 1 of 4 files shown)


Cooldude2606:
I think it would be good to include hazard concrete into this menu.
I have a suggestion that should be easy to add, when you click a colour it will attempt to place concrete into your hand, this will make using the gui much faster.
I have found that adding a second tile placement before you add the coloured concrete allows use of tile.hidden_tile which now means that floating text is not required.
Example:
```lua
surface.set_tiles{
    { name = tile.hidden_tile, position = tile.position },
    { name = 'blue-refined-concrete', position = tile.position }
}
```

bbassie:
Now after testing I remember that I have tried something similar, but it doesn't work because when in eraser mode. When you want to place concrete on top of the already placed colored concrete it places below the colored concrete removing the old tile. If you try to get the old tile in that event it will point to the colored concrete not allowing you to place back the old surface tiles.

bbassie:
Closing pull request because there does not seem to be an easy old tile solution.

PR #179, opened 2020-11-03
Filter out `blueprint` and `blueprint-book` from quick bar player data

### LINKED ISSUE #178: Toolbar blueprint book bad state
**Scenario version**
Version number: Current master

**Bug description**
When you've got a blueprint book in the toolbar and do /save-toolbar and then rejoin the toolbar wil be in a bad state:
image

bbassie:
Fixes #178


Cooldude2606 (approved):
Very small change, evenything looks good.

PR #183, opened 2020-11-06
Changed choose_elem from item type to signal

### LINKED ISSUE #181: More icons for the warp system
**Feature overview**
[11:14 PM] cabb99: can we also get fluid and logic icons?
image

bbassie:
Fixes #181

Also fixes that the warp tiles are left on the map because there is no radar coverage.


Cooldude2606 (approved):
Checked through everything, looks good and allows more symbols to be used.

PR #186, opened 2020-11-07
Remove/kill button

bbassie:


Cooldude2606:
There was no objection from staff so this will be considered, I will allow a little longer to make sure there are still no objections.

Cooldude2606:
Three people have objected to this change and we are now discussing how their concerns can be addressed.

Cooldude2606 (approved):
I have spoken with all those who opposed the removal and have come to arangements that they are happy with.

PR #187, opened 2020-11-07
Removed Temp ban related stuff

bbassie:


Cooldude2606:
There was no objection from staff so this will be considered, I will allow a little longer to make sure there are still no objections.

modules/control/jail.lua
```diff
@@ -46,37 +40,22 @@ local Jail = {
...
 Global.register({
-    old_roles = old_roles,
-    temp_bans = temp_bans
+    old_roles = old_roles
 }, function(tbl)
     Jail.old_roles = tbl.old_roles
-    Jail.temp_bans = tbl.temp_bans
     old_roles = Jail.old_roles
-    temp_bans = Jail.temp_bans
 end)
```
Cooldude2606, line 52: As `old_roles` is the only field, the outter most table can be removed.
```lua
Global.register(old_roles, function(tbl)
    Jail.old_roles = tbl
    old_roles = tbl
end)
```

Cooldude2606 (changes requested):
There was no further objections to the removal of temp ban so this pr can go a head.
Only one change is required before merging.

Cooldude2606 (approved):
Everything looks good, happy to merge pr.

PR #190, opened 2020-11-08
Feature: Warps need to be connected to the same electricity network

bbassie:
This pull request contains a refactor of the warp list to enable to make it easier to implement this feature.
image

image


bbassie:
Creating a warp will now error if you are standing too close to water, this should also fix being able to create land with warps.
image

Cooldude2606:
- The new warp looks nice, makes me wonder if the center colour could be based on the icon selected, possibility through using a table lookup.
- The left padding on the text feilds needs to be set back to 0, the text looks to cramped being pushed against the edge.
- Warp proximity bypass is not working correctly, any role which has the bypass is unable to warp at all, the expected behaviour is that anyone with the bypass is able to goto any warp without the requirement of being on a warp. I suspect the issue is because the permission is not checked for during the new connected network check.
- Longer names break the scroll bar, although I think this might be a factorio issue.

bbassie:
> * The new warp looks nice, makes me wonder if the center colour could be based on the icon selected, possibility through using a table lookup.

Maybe for the future ;)

> * The left padding on the text fields needs to be set back to 0, the text looks to cramped being pushed against the edge.

Added a bit more padding

> * Warp proximity bypass is not working correctly, any role which has the bypass is unable to warp at all, the expected behavior is that anyone with the bypass is able to goto any warp without the requirement of being on a warp. I suspect the issue is because the permission is not checked for during the new connected network check.

Proximity should be working now

> * Longer names break the scroll bar, although I think this might be a Factorio issue.

This seems to be a Factorio issue yea

Cooldude2606:
All previous comments have been adressed, I have a few further suggestions for colours and tooltips.
- The tooltips will look better if kepted short and palced onto a new line. I suggest `Go to x __1__ y __2__\nDistance Bypass` and `Go to x __1__ y __2__\nNetwork Bypass`.
- On the topic of tooltips, adding `\nCooldown Bypass` onto the progress bar if the player has the permission would keep styling consistent, this information is not currenrly present anywhere.
- Using grey for the current warp does not provide enough contrast against the white of the other names so I suggest having the current warp show as blue.
- Blue can also be used to show that a player is unable warp because of they cooldown.
- Final style suggestion, using blue for all warps when a player is not on a warp but has bypass does not make much sence to me, this will be the base state of the warps so I think it should be white.
- One coding suggestion I found while playing with the colours, your if statements in `update_icon_button` can be reduced with the following format:
```lua
if not warp_player_is_on then
    if bypass_warp_proximity then
        --[[ Block - Bypass Not In Range ]]
    else
        --[[ Block - Not In Range ]]
    end
elseif warp_player_is_on.warp_id == warp.warp_id then
    --[[ Block - Same Warp ]]
elseif on_cooldown then
    --[[ Block - Cooldown ]]
else
    if warp_electric_network_id == player_warp_electric_network_id then
        --[[ Block - Allowed ]]
    elseif bypass_warp_proximity then
        --[[ Block - Bypass Different Network ]]
    else
        --[[ Block - Different Network ]]
    end
end
```
The following shows the warp system I used to test and the colours I described above.
warpSystem
warpColours

bbassie:
@Cooldude2606 Please fix CI & review my latest commit.

Cooldude2606:
Git actions have beeen fixed, the fix has been merged into your pr.

modules/gui/warp-list.lua
```diff
@@ -307,56 +405,130 @@ Gui.element{
...
+
+-- Helper function to style and enable or disable a button element
+local function update_icon_button(element, on_cooldown, warp, warp_player_is_on, bypass_warp_proximity)
+    -- Check if button element is valid
+    if not element or not element.valid then return end
+
+    local label_style = element.parent.parent['name-'..warp.warp_id][warp_label.name].style
+    local warp_status_element = element.parent.parent['name-'..warp.warp_id][warp_status.name]
+
+    if not warp_player_is_on then
```
Cooldude2606, line 417: A comment for each case would make this more readable.

modules/gui/warp-list.lua
```diff
@@ -307,56 +405,130 @@ Gui.element{
...
-        if element and element.valid then
-            element.enabled = not button_disabled
-            if button_disabled then
-                element.tooltip = {'warp-list.goto-disabled'}
-            else
-                local position = Warps.get_warp(warp_id).position
-                element.tooltip = {'warp-list.goto-tooltip', position.x, position.y}
-            end
-        end
+    local bypass_warp_proximity = check_player_permissions(player, 'bypass_warp_proximity')
```
Cooldude2606, line 486: Again just preference so non issue, but I would have placed this next to `local warp_player_is_on` rather than between the table and iterator.

modules/gui/warp-list.lua
```diff
@@ -365,54 +537,52 @@ local function update_warp(player, warp_table, warp_id)
...
+    else
+        -- Set the icon elements visibility
+        warp_icon_element.visible = true
         -- The SpritePath type is not the same as the SignalID type
-        local sprite = warp_icon.type .. '/' ..warp_icon.name
-        if warp_icon.type == 'virtual' then
-            sprite = 'virtual-signal/' ..warp_icon.name
+        local sprite = warp.icon.type .. '/' ..warp.icon.name
+        if warp.icon.type == 'virtual' then
+                sprite = 'virtual-signal/' ..warp.icon.name
```
Cooldude2606, line 563: Double indentation.

modules/gui/warp-list.lua
```diff
@@ -365,54 +537,52 @@ local function update_warp(player, warp_table, warp_id)
...
-        warp_icon_editing(icon_flow, warp)
-
+        warp_icon_element.sprite = sprite
+        -- Set icon edit to the warps icon
+        warp_icon_edit_element.elem_value = warp.icon
+        warp_icon_edit_element.visible = false
+        -- Set the name elements visibility
+        label_element.visible = true
+        label_element.caption = warp.name
+        textfield_element.visible = false
```
Cooldude2606, line 572: When not editing it would be good UX to set the textfield text to the warp name, so when a player later switches to edit mode the text is not outdated.

Cooldude2606 (commented):
Due to the time that I am reviewing this, I have not tested in game or looked at styling.
Most of the comments are suggestions that could be considered rather than required changes.
I will review the styling tomorrow.

modules/gui/warp-list.lua
```diff
@@ -30,10 +30,18 @@ end)
 
 --- Styles used for sprite buttons
 local Styles = {
-    sprite20 = Gui.sprite_style(20),
-    sprite22 = Gui.sprite_style(20, nil, { right_margin = -3 }),
+    sprite22 = { height = 22, width = 22, padding = -2 },
```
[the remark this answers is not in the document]
bbassie: I preferred it like this 👍

modules/gui/warp-list.lua
```diff
@@ -73,87 +81,199 @@ Gui.element{
...
+--- Warp status, visible if the player is not in edit state
+--- This will show if the warp is connected or not
+-- @element warp_status
+local warp_status =
+Gui.element(function(event_trigger, parent)
+    -- Draw the element
+    return parent.add{
+        name = event_trigger,
+        type = 'label',
+        caption = '[img=utility/electricity_icon_unplugged]', -- Temporary icon
```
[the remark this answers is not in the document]
bbassie: To make it easier to know if something is wrong because someone should never see the icon.

modules/gui/warp-list.lua
bbassie: Good catch

Cooldude2606 (commented):
All suggested changes which where implemented look correct.
I will do a final check in game before I approve the PR.

Cooldude2606:
Comments from in game testing:
- The "bypass" and "same warp" feel smaller than the other icons, manual size adjustment might be required.
- The "not in range" icon feels too large compered to the other icons, manual size adjustment might be required.
- The description of "bypass" in the sub tooltip is too long, excluding it based on permissions would be a bonus.
- The info doesnt sit right, idk if its because its the only blue thing or if its the size, but something feels off with it. (it could be its new and im not used to seeing it)
- Status icons are not vertically centered, when the name wraps onto a new line the icon does not remained centered.
- Making the gui a bit wider to compensate for the room now taken up by the icon would be a plus. (maybe only when the player has edit accessed, since without edit the name can expand further right)
- Should presses "Discard changes" remove a warp with the edit was opened as a result of presing "New warp"? (i think this would require a bit of work todo and would not provide much benefit beyound consistency)

bbassie:
> Comments from in game testing:
> * The "bypass" and "same warp" feel smaller than the other icons, manual size adjustment might be required.
> * The "not in range" icon feels too large compered to the other icons, manual size adjustment might be required.

When I have the time I'll look for better icons, doing custom formatting for each icon seems too over done.

> * The description of "bypass" in the sub tooltip is too long, excluding it based on permissions would be a bonus.

Will change the description, doing the second part maybe.

> * The info doesnt sit right, idk if its because its the only blue thing or if its the size, but something feels off with it. (it could be its new and im not used to seeing it)

Think because it's new, maybe good to add to other GUI's to inform users about what it does.

> * Status icons are not vertically centered, when the name wraps onto a new line the icon does not remained centered.
> * Making the gui a bit wider to compensate for the room now taken up by the icon would be a plus. (maybe only when the player has edit accessed, since without edit the name can expand further right)

I'll make it single line and increase the with if it gets too small (like you said maybe per user, but we'll see).

> * Should presses "Discard changes" remove a warp with the edit was opened as a result of presing "New warp"? (i think this would require a bit of work todo and would not provide much benefit beyound consistency)

Will look into this, it’s interesting!

locale/en/gui.cfg
```diff
@@ -93,20 +93,34 @@ invalid=Autofill set to maximum amount: __1__ __2__ for __3__
...
+sub-tooltip-current= - __1__ This is the warp you are currently standing on;
+sub-tooltip-connected= - __1__ This warp is in the same electricity network;
+sub-tooltip-different= - __1__ This warp is not connected to the one u are standing on;
+sub-tooltip-bypass= - __1__ Because your role you are allowed to go to this warp;
+sub-tooltip-not_available= - __1__ This warp is out of range;
+sub-tooltip-cooldown= - __1__ You are currently on cooldown;
 too-close=Can't make warp; too close to warp: __1__
+too-close-to-water=Cannot create warp this close to water, please move __1__ tiles away from the water body
+too-close-to-entities=Cannot create warp this close to entities, please move __1__ tiles away from the entities or destroy them
+too-close-to-neutral-entities=Cannot create warp this close to neutral entities, please move __1__ tiles away from the neutral entities or destroy them
```
Cooldude2606, line 108: Unused locale, and I am not sure what the use case would be.

locale/en/gui.cfg
```diff
@@ -93,20 +93,34 @@ invalid=Autofill set to maximum amount: __1__ __2__ for __3__
...
 [warp-list]
-main-caption=Warp List
+main-caption=Warp List [img=info]
 main-tooltip=Warp List; Must be within __1__ tiles to use
-sub-tooltip=Warps can only be used every __1__ seconds and when within __2__ tiles
+sub-tooltip=Warps can only be used every __1__ seconds and when within __2__ tiles\n__3__\n__4__\n__5__\n__6__\n__7__\n__8__
+sub-tooltip-current= - __1__ This is the warp you are currently standing on;
+sub-tooltip-connected= - __1__ This warp is in the same electricity network;
+sub-tooltip-different= - __1__ This warp is not connected to the one u are standing on;
+sub-tooltip-bypass= - __1__ Because your role you are allowed to go to this warp;
```
Cooldude2606, line 102: Should be "Because of your role"

Cooldude2606 (commented):
- New icons look great, even if this was only a result of updating to 1.1
- Remove warp on first edit works as expected, but it might be possible to use a boolean?
- One change was missed relating to playing a sound, one was corrected and the other not.
- Because of the truncated warp name, could the full name be added at the top of the tooltip.
- I will be asking for an additional review of the locale file to ensure it is consistent.

modules/control/warps.lua
```diff
@@ -37,9 +37,16 @@ Global.register(force_warps, function(tbl)
...
+-- Create a array of entity names that will be added to the remove filter
+local remove_warp_area_entity_names = {}
+for _, entity in pairs(config.entities) do
+    table.insert(remove_warp_area_entity_names, entity[1])
+end
+
 -- When a warp is updated change its chat tag and resort the warp order
 WrapData:on_update(function(warp_id, warp, old_warp)
     if warp then
+        warp.updates = warp.updates + 1
```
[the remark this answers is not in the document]
bbassie: It's the second update that I'm looking for.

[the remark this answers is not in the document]
Cooldude2606: Oh ok well that is fine then, dont think they would be any benifit from trying to force a boolean method here.

locale/en/gui.cfg
```diff
@@ -93,20 +93,33 @@ invalid=Autofill set to maximum amount: __1__ __2__ for __3__
...
 
 [warp-list]
-main-caption=Warp List
+main-caption=Warp List [img=info]
 main-tooltip=Warp List; Must be within __1__ tiles to use
-sub-tooltip=Warps can only be used every __1__ seconds and when within __2__ tiles
+sub-tooltip=Warps can only be used every __1__ seconds and when within __2__ tiles\n__3__\n__4__\n__5__\n__6__\n__7__\n__8__
+sub-tooltip-current= - __1__ This is the warp you are currently standing on;
+sub-tooltip-connected= - __1__ This warp is in the same electricity network;
+sub-tooltip-different= - __1__ This warp is not connected to the one u are standing on;
```
mark9064, line 101: u -> you

locale/en/gui.cfg
```diff
@@ -93,20 +93,33 @@ invalid=Autofill set to maximum amount: __1__ __2__ for __3__
 inserted=Inserted __1__ __2__ into __3__
 
 [warp-list]
-main-caption=Warp List
+main-caption=Warp List [img=info]
 main-tooltip=Warp List; Must be within __1__ tiles to use
-sub-tooltip=Warps can only be used every __1__ seconds and when within __2__ tiles
+sub-tooltip=Warps can only be used every __1__ seconds and when within __2__ tiles\n__3__\n__4__\n__5__\n__6__\n__7__\n__8__
+sub-tooltip-current= - __1__ This is the warp you are currently standing on;
+sub-tooltip-connected= - __1__ This warp is in the same electricity network;
```
mark9064, line 100: Is "in" the best word to describe this relationship? Maybe "on" the same network would be better. All round this tooltip is quite vague, what exactly is it trying to express

locale/en/gui.cfg
```diff
@@ -93,20 +93,33 @@ invalid=Autofill set to maximum amount: __1__ __2__ for __3__
...
+sub-tooltip-different= - __1__ This warp is not connected to the one u are standing on;
+sub-tooltip-bypass= - __1__ This warp is possible to go to because of your role;
+sub-tooltip-not_available= - __1__ This warp is out of range;
+sub-tooltip-cooldown= - __1__ You are currently on cooldown;
```
mark9064, line 104: Maybe add a "wait n seconds"? Not really needed, and probably not worth it if difficult to implement

locale/en/gui.cfg
```diff
@@ -93,20 +93,33 @@ invalid=Autofill set to maximum amount: __1__ __2__ for __3__
...
 too-close=Can't make warp; too close to warp: __1__
+too-close-to-water=Cannot create warp this close to water, please move __1__ tiles away from the water body
+too-close-to-entities=Cannot create warp this close to entities, please move __1__ tiles away from the entities or destroy them
```
mark9064, line 107:
```suggestion
too-close-to-entities=Cannot create warp this close to other entities, please move __1__ tiles away or destroy them
```

locale/en/gui.cfg
```diff
@@ -93,20 +93,33 @@ invalid=Autofill set to maximum amount: __1__ __2__ for __3__
...
 edit-tooltip=Currently being edited by: __1__
 edit-tooltip-none=Currently being edited by: Nobody
-discard-tooltip=Remove warp
+remove-tooltip=Remove warp
 timer-tooltip=Warp cooldown takes __1__ seconds
 goto-tooltip=Go to x __1__ y __2__
-goto-disabled=You must be on a warp point and have a full charge to warp
+goto-bypass=Go to x __1__ y __2__, bypass mode
+goto-bypass-different-network=Go to x __1__ y __2__, bypass mode different network
+goto-cooldown=You are on cooldown, wait for the cooldown to recharge
```
mark9064, line 119: Again a countdown might be nice here (I don't actually know where these messages are presented, so use your better judgement)

modules/control/warps.lua
```diff
@@ -37,9 +37,16 @@ Global.register(force_warps, function(tbl)
     force_warps = tbl
 end)
 
+-- Create a array of entity names that will be added to the remove filter
```
mark9064, line 40:
```suggestion
-- Create an array of entity names that will be added to the remove filter
```

modules/control/warps.lua
```diff
@@ -37,9 +37,16 @@ Global.register(force_warps, function(tbl)
...
+local remove_warp_area_entity_names = {}
+for _, entity in pairs(config.entities) do
+    table.insert(remove_warp_area_entity_names, entity[1])
+end
+
 -- When a warp is updated change its chat tag and resort the warp order
```
mark9064, line 46:
```suggestion
-- When a warp is updated change its chat tag and restore the warp order
```
I'm assuming resort is a typo here? Also yes I see this isn't technically in the diff but close enough :P

modules/control/warps.lua
```diff
@@ -148,31 +155,15 @@ function Warps.make_warp_area(warp_id)
     local position = warp.position
     local posx = position.x
     local posy = position.y
-    local radius = config.standard_proximity_radius
-    local radius2 = radius^2
 
     -- Get the tile that is being replaced, store.update not needed as we dont want it to trigger
```
mark9064, line 159:
```suggestion
    -- Get the tile that is being replaced, store.update not needed as we don't want it to trigger
```
Same as above

modules/control/warps.lua
```diff
@@ -202,22 +198,15 @@ function Warps.remove_warp_area(warp_id)
...
-    local radius2 = radius^2
 
     -- Check that a warp area was created previously
-    local base_tile = warp.old_tile
-    if not base_tile then return end
+    local old_tile = warp.old_tile
+    if not old_tile then return end
 
-    -- Reset all the tiles that were replaced
+    -- Loop over warp tiles and set them to the old tile that was below
```
mark9064, line 206: Unsure what exactly this is doing, but would something like this be better
```suggestion
    -- Restore the original tiles before the creation of the warp
```

modules/control/warps.lua
```diff
@@ -226,9 +215,15 @@ function Warps.remove_warp_area(warp_id)
         {position.x+radius, position.y+radius}
     }
 
-    -- Remove all the entities that are in the area
-    local entities = surface.find_entities_filtered{ force='neutral', area=area }
-    for _, entity in pairs(entities) do if entity and entity.valid and entity.name ~= 'player' then entity.destroy() end end
+    -- Remove warp structure entities
+    local entities = surface.find_entities_filtered{ force='neutral', area=area, name = remove_warp_area_entity_names }
+    for _, entity in pairs(entities) do
+        -- Destroy them, this will leave corpses of the entities that it distroyed.
```
mark9064, line 221:
```suggestion
        -- Destroy them, this will leave corpses of the entities that it destroyed.
```

modules/control/warps.lua
```diff
@@ -226,9 +215,15 @@ function Warps.remove_warp_area(warp_id)
...
+        if entity and entity.valid and entity.destructible == false then
+            entity.destructible = true
+            entity.die(entity.force)
+        end
+    end
 
     -- Rechart map area, usefull if warp is not covered by a radar
```
mark9064, line 228:
```suggestion
    -- Rechart map area, useful if warp is not covered by a radar
```

modules/gui/warp-list.lua
```diff
@@ -73,87 +81,200 @@ Gui.element{
...
+    }
+end)
+:style{
+    -- when editing mode because textbox is larger the icon would move up.
```
mark9064, line 237:
```suggestion
    -- When editing mode because textbox is larger the icon would move up.
```
Generally comments are being capitalised?

modules/gui/warp-list.lua
```diff
@@ -73,87 +81,200 @@ Gui.element{
...
+    -- Draw the element
+    return parent.add{
+        name = event_trigger,
+        type = 'textfield',
+        text = warp.name,
+        clear_and_focus_on_right_click = true
+    }
+end)
+:style{
+    -- Needed fields to make it squashable and strechable.
```
mark9064, line 255:
```suggestion
    -- Required fields to make it squashable and strechable.
```

modules/gui/warp-list.lua
```diff
@@ -180,117 +301,101 @@ Gui.element{
 }
 :style(Styles.sprite22)
 :on_click(function(player, element)
-    local warp_id = element.parent.name
+    local warp_id = element.parent.caption
+    -- Check if this is this is the first edit, if so remove the warp.
```
mark9064, line 305:
```suggestion
    -- Check if this is the first edit, if so remove the warp.
```

modules/gui/warp-list.lua
```diff
@@ -180,117 +301,101 @@ Gui.element{
...
--- @element warp_icon_button
-warp_icon_button =
-Gui.element(function(event_trigger, parent, warp)
-    local warp_position = warp.position
+local update_all_warp_elements
+--- Set of three elements which make up each row of the warp table
+-- @element add_warp_elements
+local add_warp_elements =
+Gui.element(function(_, parent, warp)
+    -- Add icon flow this will contain the warp button and warp icon edit button
```
mark9064, line 349:
```suggestion
    -- Add icon flow, this will contain the warp button and warp icon edit button
```

modules/gui/warp-list.lua
```diff
@@ -180,117 +301,101 @@ Gui.element{
...
+    warp_icon_editing(icon_flow, warp)
 
-    -- Draw the element
-    return parent.add{
-        name = event_trigger,
-        type = 'sprite-button',
-        sprite = sprite,
-        tooltip = {'warp-list.goto-tooltip', warp_position.x, warp_position.y},
-        style = 'slot_button'
+    -- Add name flow this will contain the warp label and textbox
```
mark9064, line 361:
```suggestion
    -- Add name flow, this will contain the warp label and textbox
```

modules/gui/warp-list.lua
```diff
@@ -180,117 +301,101 @@ Gui.element{
...
-local warp_icon_editing =
-Gui.element(function(_, parent, warp)
-    return parent.add{
-        name = warp_icon_button.name,
-        type = 'choose-elem-button',
-        elem_type = 'signal',
-        signal = {type = warp.icon.type, name = warp.icon.name},
-        tooltip = {'warp-list.goto-edit'}
+
+    -- Add button flow this will contain buttons to manage this specific warp
```
mark9064, line 375:
```suggestion
    -- Add button flow, this will contain buttons to manage this specific warp
```

modules/gui/warp-list.lua
```diff
@@ -307,56 +412,138 @@ Gui.element{
...
+        local warp_electric_network_id = warp.electric_pole and warp.electric_pole.electric_network_id or -1
+        local player_warp_electric_network_id = warp_player_is_on.electric_pole and warp_player_is_on.electric_pole.electric_network_id or -2
+        if warp_electric_network_id == player_warp_electric_network_id then
+            local position = warp.position
+            element.tooltip = {'warp-list.goto-tooltip', position.x, position.y}
+            element.enabled = true
+            warp_status_element.tooltip = {'warp-list.goto-tooltip', position.x, position.y}
+            warp_status_element.caption = warp_status_icons.connected
+            label_style.font = 'default-semibold'
+        -- If the warp is not in the same network but the player is allowed to warp without being on a warp
```
mark9064, line 466: Again "in" might not be the best but probably irrelevant for code comments

modules/gui/warp-list.lua
```diff
@@ -307,56 +412,138 @@ Gui.element{
...
+        update_warp_elements(element, next_warp, warp_player_is_on, on_cooldown, bypass_warp_proximity)
     end
 end
 
 --- Updates a warp for a player
 local function update_warp(player, warp_table, warp_id)
     local warp = Warps.get_warp(warp_id)
 
-    -- Warp no longer exists so should be removed from the list
+    -- If the warp does no longer exist then remove the warp elements from the warp table
```
mark9064, line 513:
```suggestion
    -- If the warp no longer exists then remove the warp elements from the warp table
```

modules/gui/warp-list.lua
```diff
@@ -307,56 +412,138 @@ Gui.element{
...
     if not warp then
-        remove_warp_base(warp_table, warp_id)
+        remove_warp_elements(warp_table, warp_id)
         return
     end
 
-    -- Get the warp flow for this warp
-    local warp_flow = warp_table[warp_id] or add_warp_base(warp_table, warp_id)
+    -- Check if the warp elements are created, if they are not then create a new set of them
```
mark9064, line 519:
```suggestion
    -- Create the warp elements if they do not already exist
```

modules/gui/warp-list.lua
```diff
@@ -307,56 +412,138 @@ Gui.element{
...
 
-    -- Update the edit flow
-    local edit_flow = warp_table['edit-'..warp_id]
+    -- Hide the edit button if the player is not allowed to edit the warp
     local player_allowed_edit = check_player_permissions(player, 'allow_edit_warp', warp)
     local players_editing = table.get_keys(warp.currently_editing)
-    local edit_warp_element = edit_flow[edit_warp.name]
-    local discard_warp_element = edit_flow[discard_warp.name]
-
+    -- button_flow.visible = player_allowed_edit
```
mark9064, line 543: This appears to be a commented line of code, should this be present?

modules/gui/warp-list.lua
```diff
@@ -365,54 +552,53 @@ local function update_warp(player, warp_table, warp_id)
         edit_warp_element.tooltip = {'warp-list.edit-tooltip-none'}
     end
 
-    -- Check if the player is was editing and/or currently editing
-    local warp_label_element = warp_flow[warp_label.name] or warp_label(warp_flow, warp)
-    local icon_entry = icon_flow[warp_icon_button.name] or warp_icon_button(icon_flow, warp)
-    local player_was_editing = icon_entry.type == 'choose-elem-button'
+    -- Set the warp elements visibility based on if the user is editing or not
```
mark9064, line 555:
```suggestion
    -- Set the visibility of the warp elements based on whether the user is editing or not
```

modules/gui/warp-list.lua
```diff
@@ -365,54 +552,53 @@ local function update_warp(player, warp_table, warp_id)
...
-    if not player_was_editing and not player_is_editing then
-        -- Update the warp name label and icon
-        local warp_name = warp.name
-        local warp_icon = warp.icon
-        local last_edit_name = warp.last_edit_name
-        local last_edit_time = warp.last_edit_time
-        warp_label_element.caption = warp_name
-        warp_label_element.tooltip = {'warp-list.last-edit', last_edit_name, format_time(last_edit_time)}
+    if player_is_editing then
+        -- Set the icon elements visibility
```
mark9064, line 558: There should be possessive apostrophes here, but somehow I doubt that you care :P

modules/gui/warp-list.lua
```diff
@@ -647,13 +850,18 @@ local function role_update_event(event)
     local player = game.players[event.player_index]
     local container = Gui.get_left_element(player, warp_list_container).container
 
+    -- Check if user has permission to add warps
+    local allow_add_warp = check_player_permissions(player, 'allow_add_warp')
+    -- Update container size depending on if the player is allowed to add warps
```
mark9064, line 855:
```suggestion
    -- Update container size depending on whether the player is allowed to add warps
```
This grammar error seems to be in quite a few places so I'll let you decide whether :wink: you fancy fixing it

mark9064 (commented):
I might've missed a few things as always. I also haven't checked any variable names for errors. Looks pretty good overall though!

locale/en/gui.cfg
Cooldude2606: Re-worded these five statements to be shorter and more direct.

locale/en/gui.cfg
Cooldude2606: Main issue with this request having an extra element which updates every tick; further more, this is a tooltip that will not be looked at more than a few times once players learn the symbols. Therefore, the small cost to preformance already out weights the small gain in utility.

locale/en/gui.cfg
Cooldude2606: Made requested change and changed `too-close` to use "Cannnot create warp" rather than "Can't make warp"

locale/en/gui.cfg
Cooldude2606: This tooltip shows on the warp button while it is disabled. It would be simple to add the cooldown to the tooltip; however, it may be misleading as the tooltips are only changed when entering/leaving a warp or when your cooldown hits 0. Therefore, I will not be adding the countdown to the button tooltip; however I will add the countdown onto the progress bar which alreadey updates every tick.

modules/gui/warp-list.lua
Cooldude2606: Changed to "on"

modules/gui/warp-list.lua
Cooldude2606: Removed commented code. This code was used to hide all buttons, but this required extra logic for the table and required a full redraw for permission changes; the new method does not require this.

modules/gui/warp-list.lua
Cooldude2606: Correcting spellings is helpful, but who uses apostrophes in comments XP
No change made.

Cooldude2606 (approved):
My last review covered all functional code, and marks review covered all spelling and grammar, with my last changes which implimented marks request, I am happy to merge this pr.

PR #191, opened 2020-11-08
Corpse light and line addition

bbassie:
image

### DIFF
--- a/modules/addons/death-logger.lua
+++ b/modules/addons/death-logger.lua
@@ -77,7 +77,45 @@ Event.add(defines.events.on_player_died, function(event)
         create_map_tag(death)
     end
     table.insert(deaths, death)
+
+    -- Draw a light attached to the corpse with the player color
+    if config.show_light_at_corpse then
+        rendering.draw_light{
+            sprite = 'utility/light_medium',
+            color = player.color,
+            target = corpse,
+            force = player.force,
+            surface = player.surface
+        }
+    end
 end)
+
+-- Draw lines to the player corpse
+if config.show_line_to_corpse then
+    Event.add(defines.events.on_player_respawned, function(event)
+        local player = game.players[event.player_index]
+
+        for _, death in pairs(deaths) do
+            -- Check if the death body is from the player
+            -- Check if the corpse entity is still valid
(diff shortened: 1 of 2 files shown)


modules/addons/death-logger.lua
```diff
@@ -77,8 +77,46 @@ Event.add(defines.events.on_player_died, function(event)
...
+        }
+    end
 end)
 
+-- Draw lines to the player corpse
+if config.show_line_to_corpse then
+    Event.add(defines.events.on_player_respawned, function(event)
+        local player = game.players[event.player_index]
+
+        for _, death in pairs(deaths) do
```
Cooldude2606, line 98: Due to table.insert being used the newest death will always be at the end of the array.
It will save alot of iterations by starting at the end of the array and counting backwards.
`for index = #deaths, 1, -1 do`

modules/addons/death-logger.lua
```diff
@@ -77,8 +77,46 @@ Event.add(defines.events.on_player_died, function(event)
...
+            -- Check if the death body is from the player
+            -- Check if the corpse entity is still valid
+            if death.player_name == player.name and death.corpse and death.corpse.valid then
```
Cooldude2606, line 101: Starting from the end of the list means you get newest to oldest death entries.
If you only want the latest death then I would break as soon as a matching name is found.
If you want all valid corpses then I would add a break condidition which checks `time_of_death` against the current game tick. If it is longer than 15 minutes for any death entry then you know all older corpses will have expired, and it is safe to exit the loop.
Based on above add either a break for a single corpse or a break based on `time_of_death`

Cooldude2606 (changes requested):
Nice feature, a way to hide all tracers would be nice but is not required.
There are two optimizations which should be made.

modules/addons/death-logger.lua
bbassie: Fixed

modules/addons/death-logger.lua
bbassie: Added

Cooldude2606 (approved):
Everything looks good and requested changes have been implemented.

PR #195, opened 2021-01-24
Tasklist redesign

bbassie:
* Removed the ability to pass `task_number` to `Task.add_task`
* Unified button styles of the Gui
* Made task-list code simular to the new warp-list


Cooldude2606:
Git actions have beeen fixed, the fix has been merged into your pr.

bbassie:
Implemented a base working version of the task list with a new redesign.

bbassie:
@Cooldude2606 I don't know what else to add to this, so it should be ready for review if you have the time!

Cooldude2606:
A few small changes have been discussed which can be implemented in this PR with an option for another rewrite for the future.
Changes require for this PR:
- Although the thin buttons look nice, the icons are cropped as a result and so the buttons should be made bigger again.
- Changing the text field to a text box will allow for multiline formatting at the cost of on_confirmed. But we find this is acceptable and will allow for a "title" for a task which is defined as the first line of the message. (Similar to git messages)
- The confirm edit button should have a tooltip for why it is disabled.

Change for future redesigns:
- Rewrite of the control module to allow for easy use of a box list element rather than the custom hack which has been created here. This could include new functions for getting a task based on force and index (rather than id) and events for tasks being added, removed, and edited (rather than just updated).

Once the required UI changes are made I will review the code.

modules/gui/task-list.lua
```diff
@@ -306,6 +315,29 @@ local task_view_footer =
...
+        return message
+    end
+    -- If message has multiple lines
+    for key, value in string.gmatch(trimmed, message_pattern) do
+        message.title = key
+        message.body = value
+        break
+    end
+    return message
+end
```
bbassie: @Cooldude2606 Maybe you have a better idea on how to do this...

bbassie:
Implemented the changes requested. Lint is failing on for loop 333, maybe you know something better.

modules/gui/task-list.lua
Cooldude2606: You can use string.match rather than string.gmatch. Here is an example:
`local d, m, y = string.match("14/04/2020", "(%d+)/(%d+)/(%d+)")`

modules/gui/task-list.lua
```diff
@@ -57,360 +76,657 @@ local function check_player_permissions(player, task)
...
+            caption = caption
+        }
+    end
+)
+
+--- Action flow that contains action buttons
+-- @element subfooter_actions
+local subfooter_actions =
+    Gui.element(
+    function(_, parent)
```
Cooldude2606, line 168: Due to how close this is to the trivial case, I feel the GUI core should be changed to allow a custom non unique name.
This would be an easy change to make in core, but may require a `id` property depending on the value I want `name` to hold.
I will make this feature request, and might implement it for the next release.

modules/gui/task-list.lua
```diff
@@ -57,360 +76,657 @@ local function check_player_permissions(player, task)
...
+            name = "scroll",
+            type = "scroll-pane",
+            direction = "vertical",
+            horizontal_scroll_policy = "never",
+            vertical_scroll_policy = "auto",
+            style = "scroll_pane_under_subheader"
+        }
+        scroll_pane.style.horizontally_stretchable = true
+        scroll_pane.style.padding = 0
+        scroll_pane.style.maximal_height = 300
```
Cooldude2606, line 222: This max height should be adjusted to contain a whole number of tasks. Currently it contains 10 with some of the 11th showing.

modules/gui/task-list.lua
```diff
@@ -57,360 +76,657 @@ local function check_player_permissions(player, task)
...
+            caption = "New task"
+        }
+        title_label.style.padding = 4
+        title_label.style.font = "default-bold"
+        title_label.style.single_line = false
+        local body_label =
+            footer.add {
+            type = "label",
+            name = "body",
+            caption = "Do x or y"
```
Cooldude2606, line 305: Is this caption needed when it always gets overriden?

modules/gui/task-list.lua
```diff
@@ -57,360 +76,657 @@ local function check_player_permissions(player, task)
...
+    end
+)
+
+local message_pattern = "(.-)\n(.*)"
+
+--- Parce a string into a message object with title and body
+-- @tparam string str message data
+local function parse_message(str)
+    -- Trimm the spaces of the string
+    local trimmed = (string.gsub(str, "^%s*(.-)%s*$", "%1"))
```
Cooldude2606, line 324: I dont think the outter most brackets are required.

modules/gui/task-list.lua
```diff
@@ -57,360 +76,657 @@ local function check_player_permissions(player, task)
...
+)
+
+--- Frame element with the right styling
+-- @element subfooter_frame
+local subfooter_frame =
+    Gui.element(
+    function(_, parent, name)
+        return parent.add {
+            type = "frame",
+            name = name,
```
Cooldude2606, line 137: Because of how common cases like this are, I feel I should allow the core to generate these.
Perhaps with a `name = Gui.argument_value` in the define table, and extent this to work for any property like captions.
I will add this as a feature request.

Cooldude2606 (changes requested):
Mostly comments with small adjustments needed.
Considering the amout code there was very few errors, and I found some useful features I can add to GUI core.
There is one point I would like to discuss, because task details can vary in size I personally feel the buttons should be moved to the top right of "Task details". This would make them be in a consistent place regardless on the size of the task. However, the confirm and discard buttons are still great at the bottom as you select them last.

Cooldude2606 (approved):
All highlighted issues were resolved and the code is ready to merge.
However, I will still like to discuss the best location for "edit, delete, close" before we merge.
Also I will request a review from mark to check the locale files and comments.

bbassie:
> All highlighted issues were resolved and the code is ready to merge.
> However, I will still like to discuss the best location for "edit, delete, close" before we merge.
> Also I will request a review from mark to check the locale files and comments.

Best I can come up with:
image
It's still a bit meh but maybe better than 3 buttons in a row?

Cooldude2606:
> It's still a bit meh but maybe better than 3 buttons in a row?

Yeah it is the three buttons that is off putting for me, since it seams very cramped.
I think it looks good having the two buttons, close will probably be the least used so its ok for it to have less contrast.
(least used because people will leave a task open, then either switch to a different one or delete it when finished)
If you also think it looks better than the three button style, then I would go for it; if not we should ask others for an opinion.

Cooldude2606:
Changes work as intented; however, your `vertical_flow` is not vertical and if it was vertical then it would break styling. I would name it `horizontal_flow` `title_flow` or just `flow`.

modules/gui/task-list.lua
```diff
@@ -57,360 +76,657 @@ local function check_player_permissions(player, task)
...
+    end
+):on_click(
+    function(player, element, _)
+        local task_id = element.parent.caption
+        PlayerSelected:set(player, task_id)
+    end
+)
+
+--- Scrollable list of all tasks
+-- @element no_tasks_found
```
mark9064, line 207: `@element` is incorrect?

modules/gui/task-list.lua
```diff
@@ -57,360 +76,657 @@ local function check_player_permissions(player, task)
...
-    local task_id = element.parent.name
-    local new_message = element.parent[task_editing.name].text
-    Tasks.set_editing(task_id, player.name)
-    Tasks.update_task(task_id, new_message, player.name)
-end)
-
---- Button displayed next to tasks which the user is currently editing, used to discard changes
--- @element cancel_edit
-local cancel_edit =
+--- Button to close the view task footer
```
mark9064, line 254: `view task footer` or `task view footer`, pick one
Module wide

modules/gui/task-list.lua
```diff
@@ -57,360 +76,657 @@ local function check_player_permissions(player, task)
...
+        local action_flow = subfooter_actions(footer)
+        view_task_delete_button(action_flow)
+        view_task_edit_button(action_flow)
+        return footer
+    end
+)
+
+local message_pattern = "(.-)\n(.*)"
+
+--- Parce a string into a message object with title and body
```
mark9064, line 320:
```suggestion
--- Parse a string into a message object with title and body
```

modules/gui/task-list.lua
```diff
@@ -57,360 +76,657 @@ local function check_player_permissions(player, task)
...
+-- @tparam string str message data
+local function parse_message(str)
+    -- Trimm the spaces of the string
```
mark9064, line 323:
```suggestion
    -- Trim the spaces of the string
```

modules/gui/task-list.lua
```diff
@@ -57,360 +76,657 @@ local function check_player_permissions(player, task)
...
+    local trimmed = string.gsub(str, "^%s*(.-)%s*$", "%1")
+    local message = { title = "", body = "" }
+    local title, body = string.match(trimmed, message_pattern)
+    if not title then
+        -- If it doesn't match the patter return the str as a title
```
mark9064, line 328:
```suggestion
        -- If it doesn't match the pattern return the str as a title
```
str is one of the arguments, which specific string is this referring to (at least to me str is often used as a generic type for strings)

modules/gui/task-list.lua
```diff
@@ -57,360 +76,657 @@ local function check_player_permissions(player, task)
...
+    function(player, element, _)
+        local message = element.parent.parent[task_message_textfield.name].text
+        PlayerIsCreating:set(player, false)
+        local parsed = parse_message(message)
+        local task_id = Tasks.add_task(player.force.name, player.name, parsed.title, parsed.body)
+        PlayerSelected:set(player, task_id)
+    end
+)
+
+--- Button to discard the changes
```
mark9064, line 441: Since L387 exists, it'd be nice if these docstrings were slightly more specific. Probably minor though
Applies in a few other places actually, like subfooter in tasklist container

modules/gui/task-list.lua
```diff
@@ -57,360 +76,657 @@ local function check_player_permissions(player, task)
...
+        -- Draw task list element
+        task_list(container)
+
+        local task_view_footer_element = task_view_footer(container)
+        local task_edit_footer_element = task_edit_footer(container)
+        local task_create_footer_element = task_create_footer(container)
+        task_view_footer_element.visible = false
+        task_edit_footer_element.visible = false
+        task_create_footer_element.visible = false
+        -- Return the exteral container
```
mark9064, line 504:
```suggestion
        -- Return the external container
```

modules/gui/task-list.lua
```diff
@@ -57,360 +76,657 @@ local function check_player_permissions(player, task)
...
+    end
+):add_to_left_flow(
+    function(player)
+        local task_ids = Tasks.get_force_task_ids(player.force.name)
+        return #task_ids > 0
+    end
+)
+
+--- Button on the top flow used to toggle the task list container
+-- @element toggle_left_element
```
mark9064, line 515: I don't actually know what `@element` signifies but this one also doesn't seem correct

modules/gui/task-list.lua
```diff
@@ -57,360 +76,657 @@ local function check_player_permissions(player, task)
...
+    edit_button_element.visible = has_permission
+    delete_button_element.visible = has_permission
+    title_element.caption = task.title
+    body_element.caption = task.body
+
+    local players_editing = table.get_keys(task.currently_editing)
     if #players_editing > 0 then
-        edit_task_element.hovered_sprite = 'utility/warning_icon'
-        edit_task_element.tooltip = {'task-list.edit-tooltip', table.concat(players_editing, ', ')}
+        -- edit_task_element.hovered_sprite = 'utility/warning_icon'
```
mark9064, line 584: Dead code? (same as line a couple below)

modules/gui/task-list.lua
```diff
@@ -57,360 +76,657 @@ local function check_player_permissions(player, task)
...
-    -- Draw the no tasks label
-    local no_tasks_label =
-    scroll_table.parent.add{
-        name = 'no_tasks',
-        type = 'label',
-        caption = {'task-list.no-tasks'}
-    }
+        -- Update the task for all the players on the force
+        for _, player in pairs(force.connected_players) do
+            -- Update the task view if the player has the changed task selected
```
mark9064, line 605: Maybe better phrased as `changed the(ir) selected task`? Not 100% sure what's going on here so ignore if I've misunderstood

modules/gui/task-list.lua
```diff
@@ -57,360 +76,657 @@ local function check_player_permissions(player, task)
...
 
-        -- Update the task that was changed
-        update_task(player, scroll_table, task_id)
+        local frame = Gui.get_left_element(player, task_list_container)
+        local task_list_element = frame.container.scroll.task_list
+        local view_flow = frame.container.view
+        local edit_flow = frame.container.edit
+        local isEditing = PlayerIsEditing:get(player)
+
+        -- If the selection has a old state re-enable the button list element
```
mark9064, line 654:
```suggestion
        -- If the selection has an old state re-enable the button list element
```

mark9064: Or maybe `previous state` if more appropriate? (time vs sequence)

modules/gui/task-list.lua
```diff
@@ -57,360 +76,657 @@ local function check_player_permissions(player, task)
...
+                Tasks.set_editing(old_state, player.name, nil)
+                Tasks.set_editing(new_state, player.name, true)
+                view_flow.visible = false
+                edit_flow.visible = true
+            else
+                view_flow.visible = true
+                edit_flow.visible = false
             end
+        else
+            -- If new state nil then hide footer elements and set editing to nil for old_state
```
mark9064, line 678:
```suggestion
            -- If new state is nil then hide footer elements and set editing to nil for old_state
```

modules/gui/task-list.lua
```diff
@@ -57,360 +76,657 @@ local function check_player_permissions(player, task)
...
+        -- Set selected to the same value, this will make sure if player was in a editing state he will leave it if no permissions.
+        PlayerSelected:set(selected)
+    end
 
-    -- Update the tasks, incase the user can now edit them
-    local scroll_table = container.scroll.table
-    update_all_tasks(player, scroll_table)
-
-    -- Update the new task button incase the user can now add them
+    -- Update the new task button and create footer incase the user can now add them
```
mark9064, line 721:
```suggestion
    -- Update the new task button and create footer in case the user can now add them
```

modules/gui/task-list.lua
```diff
@@ -57,360 +76,657 @@ local function check_player_permissions(player, task)
...
 
 --- Makes sure the right buttons are present when roles change
 local function role_update_event(event)
     local player = game.players[event.player_index]
     local container = Gui.get_left_element(player, task_list_container).container
+    -- Update the view task
+    local selected = PlayerSelected:get(player)
+    if selected then
+        update_task_view_footer(player, selected)
+        -- Set selected to the same value, this will make sure if player was in a editing state he will leave it if no permissions.
```
mark9064, line 717: I'm not a fan of how this is phrased. Maybe something like `Set selected to the same value, this will make sure if the player was in a editing state they will leave it if <they no longer have permission to edit?>`(I'm not totally sure what's happening here so amend as needed)

mark9064 (approved):
Large PR so quite a few comments, but it's just a lot of little things :smile:

.github/workflows/dev-deploy.yml
```diff
@@ -18,10 +18,10 @@ jobs:
           fetch-depth: 0
 
       - name: Install Lua
-        uses: leafo/gh-actions-lua@v5
+        uses: leafo/gh-actions-lua@v8.0.0
```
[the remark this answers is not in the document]
bbassie: Because the GitHub actions stopped working and @Cooldude2606 force pushed the changes or something like that

modules/gui/task-list.lua
bbassie: @Cooldude2606 I was also looking into this, but maybe you can clarify?

modules/gui/task-list.lua
bbassie: Ignored, basically this is in a loop where it loops over all the players. If a player has the task that is being updated selected it will update the task view elements for them.

Updated the comment to hopefully be more specific in a later commit.

modules/gui/task-list.lua
bbassie: Changed this and some other occurrences of old and new

.github/workflows/dev-deploy.yml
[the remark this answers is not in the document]
Cooldude2606: Yes this change was to correct CI issues, the changes had to be merged here for the CI to pass, all other PRs were created after the change was commited to dev.

modules/gui/task-list.lua
Cooldude2606: The tag `@element` is a cosmetic feature on the docs used to signify that the module adds gui elements.
As such the name is also cosmetic, but when possible should match the variable name.
In this case there is no variable so the name describes what it does.
Perhaps `task_list_toggle` would be better.
Also there are many other cases where `toggle_left_element` is used due to neglect.
Example from previous doc generation:
docs

Cooldude2606: Due to this issue existing across multiple files, it will be addressed on a seperate pull request.

Issue #197, opened 2021-02-09
Ping last location of player


7h3w1z4rd:
**Feature overview**
Name: Ping last location of player
TL;DR: Command that gps pings last location of specific player

**Details**
Command that would ping the position of the specific player provided in the command execution. Would provide the position of the player via a gps ping in the game chat. The ping itself would only be visible to the player who executed the command.

**Possible methods**
Get player via game.player list and fetch their position.

Cooldude2606:
PR implementing this request has been merged into dev. Clossing issue.

PR #198, opened 2021-02-09
Feature/cmd last command

bbassie:
Implements #197 

Image

### DIFF
--- a/modules/commands/last-location.lua
+++ b/modules/commands/last-location.lua
@@ -11,8 +11,8 @@ require 'config.expcore.command_general_parse'
 -- @command last-location
 -- @tparam LuaPlayer player the player that you want a location of
 Commands.new_command('last-location', 'Sends you the last location of a player')
-:add_param('player', false, 'player-role')
-:set_flag('admin_only')
+:add_alias('location')
+:add_param('player', false, 'player')
 :register(function(player, action_player)
     local action_player_name_color = format_chat_player_name(action_player)
     player.print{'expcom-lastlocation.response', action_player_name_color, action_player.position.x, action_player.position.y}
--- a/modules/commands/last-location.lua
+++ b/modules/commands/last-location.lua
@@ -1,0 +1,19 @@
+--[[-- Commands Module - Last location
+    - Adds a command that will return the last location of a player
+    @commands LastLocation
+]]
+
+local Commands = require 'expcore.commands' --- @dep expcore.commands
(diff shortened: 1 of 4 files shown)


modules/commands/last-location.lua
```diff
@@ -0,0 +1,19 @@
+--[[-- Commands Module - Last location
+    - Adds a command that will return the last location of a player
+    @commands LastLocation
+]]
+
+local Commands = require 'expcore.commands' --- @dep expcore.commands
+local format_chat_player_name = _C.format_chat_player_name --- @dep expcore.common
+require 'config.expcore.command_general_parse'
```
Cooldude2606, line 8: `player-role` is included in `command_role_parse` rather than `command_general_parse`. Although I would suggest using `player` over `player-role` as I see no harm in getting the last location of a player of higher role.

modules/commands/last-location.lua
```diff
@@ -0,0 +1,19 @@
...
+
+--- Get the last location of a player.
+-- @command last-location
+-- @tparam LuaPlayer player the player that you want a location of
+Commands.new_command('last-location', 'Sends you the last location of a player')
+:add_param('player', false, 'player-role')
```
Cooldude2606, line 14: Could be `player` rather than `player-role`? See above comment.

modules/commands/last-location.lua
```diff
@@ -0,0 +1,19 @@
...
+:set_flag('admin_only')
```
Cooldude2606, line 15: `admin_only` will prevent non game admins from using the command. All auth handlers must be passed to allow command execution, since member role does not have game admin, this prevents members from using the command even when included as a permission.

Cooldude2606 (changes requested):
One issue with `admin_only` and one suggestion about `player-role` and its require.
Edit: Some rounding of the position wouldnt harm anything :)

modules/commands/last-location.lua
Cooldude2606: Now uses player rather than player-role

modules/commands/last-location.lua
Cooldude2606: Now uses player rather than player-role

modules/commands/last-location.lua
Cooldude2606: Removed flag for admin only

Cooldude2606 (approved):
I have added the changes which I requested so this pr can be merged for next release.

PR #199, opened 2021-02-13
Roles over RCON, clear join message, quickbar fix

oof2win2:
## Allow roles to be set over RCON
Since a lot of stuff is already done with RCON, why not allow roles to be set this way too, using `/assign` and `/unassign`?
Changes the player name to be either the player's name or `<server>`, as that is the only valid option for what could happen.
## Command to clear join message
New command to clear your join message. 
## Save-quickbar bugfix
The quickbar sync had an issue where items with unique IDs would be just replaced by general items, i.e. a deconstruction planner with a specific filter would be replaced by a generic deconstruction planner. This has been fixed by checking if an item is a list of set items, which include:
- Spidertron remotes (unique ID to spidertron, changes every time spidertron dies, etc)
- Deconstruction planner
- Upgrade planner
- Blueprint book
- Blueprint
If the item is in this list, it is ignored and not pushed to the user's settings of quickbar syncing.

### DIFF
--- a/config/expcore/roles.lua
+++ b/config/expcore/roles.lua
@@ -155,7 +155,8 @@ Roles.new_role('Supporter','Sup')
 :allow{
     'command/jail',
     'command/unjail',
-    'command/join-message'
+    'command/join-message',
+    'command/join-message-clear'
 }
 
 Roles.new_role('Partner','Part')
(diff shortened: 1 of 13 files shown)


oof2win2:
Whoops, merging into master. Not what I want

PR #200, opened 2021-02-13
Clear join message & fixing saving quickbar

oof2win2:
Same as #199 but not merging to master & removed my method of assigning roles over RCON (`/interface` should be used instead)
## Command to clear join message
New command to clear your join message (`/join-message-clear`)
## Save-quickbar bugfix
The quickbar sync had an issue where items with unique IDs would be just replaced by general items, i.e. a deconstruction planner with a specific filter would be replaced by a generic deconstruction planner. This has been fixed by checking if an item is a list of set items, which include:
- Spidertron remotes (unique ID to spidertron, changes every time spidertron dies, etc)
- Deconstruction planner
- Upgrade planner
- Blueprint book
- Blueprint
If the item is in this list, it is ignored and not pushed to the user's settings of quickbar syncing.
Extended version of d65a1fe0fefec17dd372860d192ac3300fbe72d8


bbassie (approved):

modules/data/greetings.lua
```diff
@@ -33,4 +33,11 @@ Commands.new_command('join-message', 'Sets your custom join message')
     if not player then return end
     CustomMessages:set(player, message)
     return {'join-message.message-set'}
+end)
+
+Commands.new_command('join-message-clear', 'Clear your join message')
+:register(function(player, message)
```
bbassie, line 39: Just remove message argument.

bbassie (changes requested):

bbassie (approved):

modules/data/quickbar.lua
```diff
@@ -37,11 +37,28 @@ Commands.new_command('save-quickbar', 'Saves your Quickbar preset items to file'
...
+        "spidertron-remote",
+        "upgrade-planner"
+    }
+    local function contains(list, x)
+        for _, v in pairs(list) do
+            if v == x then return true end
+        end
+        return false
+    end
+
```
Cooldude2606, line 53: - A constant table like this should be defined outside side the function call, this will improve preformance by removing the need to allocate memory for the table.
- Using a hash table is much faster than iteration becuase of of the seperation between C and Lua. I would remove the contains function and using a hash table in the following form:
```lua
-- Hash table
local ignoredItems = {
    ["blueprint"] = true
}

-- Test if ignored
local ignored = ignoredItems[slot.name]
```

Cooldude2606 (changes requested):
Two changes which can be made to increase preformance of the code, always define static tables outside of functions and use hash tables when possible.

Cooldude2606 (approved):
Required changes were made.

PR #203, opened 2021-04-05
Fix of player not being valid when killing themselves

oof2win2:
When a player killed themselves with PvP enabled, the game couldn't access the player. This fixes it.


Cooldude2606 (approved):
Simple change, happy to merge PR to cover your use case.

PR #204, opened 2021-04-05
Auto jail for reports and other fixes

Cooldude2606:
Reports will automatically jail a player after the combined player time of reporters exceeds 150% of the reported player's playtime. Also fixes empty reason strings, double output for admins and self reporting.

### DIFF
--- a/locale/en/addons.cfg
+++ b/locale/en/addons.cfg
@@ -76,3 +76,6 @@ get-mead-2= Skål!
 get-beer-1= 🍺 Pouring A Glass 🍺
 get-beer-2= 🍻 Chears Mate 🍻
 verify=Please return to our discord and type r!verify __1__
+
+[report-jail]
+jail=__1__ was jailed because they have been reported too many times.
--- a/locale/en/addons.cfg
+++ b/locale/en/addons.cfg
@@ -78,4 +78,4 @@ get-beer-2= 🍻 Chears Mate 🍻
 verify=Please return to our discord and type r!verify __1__
 
 [report-jail]
-jail=__1__ was jailed because they have been reported too many times.
+jail=__1__ was jailed because they were reported too many times.
--- a/modules/addons/report-jail.lua
+++ b/modules/addons/report-jail.lua
@@ -1,4 +1,4 @@
---- When a player is reported by players with a greater combined delta playtime, the player is automatically jailed
+--- When a player is reported, the player is automatically jailed if the combined playtime of the reporters exceeds the reported player
 -- @addon report-jail
 
 local Event = require 'utils.event' ---@dep utils.event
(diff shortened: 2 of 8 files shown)


modules/control/reports.lua
```diff
@@ -132,7 +132,7 @@ function Reports.report_player(player, by_player_name, reason)
     if not player then return end
     local player_name = player.name
 
-    reason = reason or 'Non given.'
+    if reason == nil or not reason:find("/S") then reason = 'Non given' end
```
mark9064, line 135: Non -> None

modules/gui/player-list.lua
```diff
@@ -72,9 +72,10 @@ Gui.element{
 }
 :style(Gui.sprite_style(30, -1, { left_margin = -2, right_margin = -1 }))
 :on_click(function(player, element)
-    local reason = element.parent.entry.text or 'Non Given'
+    local reason = element.parent.entry.text
     local action_name = SelectedAction:get(player)
     local reason_callback = config.buttons[action_name].reason_callback
+    if reason == nil or not reason:find("/S") then reason = 'no reason given' end
```
mark9064, line 78: I think `no reason given` is better than `none given`, perhaps `reports.lua` should be updated to use the same string.

modules/addons/report-jail.lua
```diff
@@ -0,0 +1,25 @@
+--- When a player is reported by players with a greater combined delta playtime, the player is automatically jailed
+-- @addon report-jail
+
+local Event = require 'utils.event' ---@dep utils.event
+local Jail = require 'modules.control.jail' ---@dep modules.control.jail
+local Reports = require 'modules.control.reports' --- @dep modules.control.reports
+local format_chat_player_name = _C.format_chat_player_name --- @dep expcore.common
+
+--- Returns the playtime of the reporter, used to sum the playtime of all reporters
```
mark9064, line 9: If I'm reading this correctly, this docstring contains a description and the reason for this function existing. This seems a bit unusual given that most other docstrings seem to describe the function only. If you would like to keep the reason, I think it could be better expressed as `Returns the playtime of the reporter. Used when calculating the total playtime of all reporters` or similar

locale/en/addons.cfg
```diff
@@ -75,4 +75,7 @@ get-mead-1= Filling the drinking horn
 get-mead-2= Skål!
 get-beer-1= 🍺 Pouring A Glass 🍺
 get-beer-2= 🍻 Chears Mate 🍻
-verify=Please return to our discord and type r!verify __1__
\ No newline at end of file
+verify=Please return to our discord and type r!verify __1__
+
+[report-jail]
+jail=__1__ was jailed because they have been reported too many times.
```
mark9064, line 81: have been -> were
The current version is grammatically correct, but I think were better expresses the situation

modules/addons/report-jail.lua
```diff
@@ -0,0 +1,25 @@
...
+local function reporter_playtime(_, by_player_name, _)
+    local player = game.get_player(by_player_name)
+    if player == nil then return 0 end
+    return player.online_time
+end
+
+--- Tests the combined playtime of all reporters against the reported player
```
mark9064, line 16: If this is the same operation as the module docstring, I think this expresses the logic better

mark9064 (approved):
Few minor things, looks good overall

modules/control/reports.lua
Cooldude2606: Used `No reaason given` so comment is outdated.

modules/gui/player-list.lua
Cooldude2606: Used this suggestion.

modules/addons/report-jail.lua
Cooldude2606: Updated the module comment to match this.

PR #206, opened 2021-04-07
Inventory Search Commands

Cooldude2606:
Added four commands which can be used to search the inventories of all players. The commands will display the top five players following 3 sort functions (the fouth command limits the 3rd to online only). These functions are based on: item count, join time, item count with play time.

### DIFF
--- a/modules/commands/search.lua
+++ b/modules/commands/search.lua
@@ -54,11 +54,11 @@ local function sort_players(players, func)
     -- Loop over all provided players
     for index, player in ipairs(players) do
         local value = func(player)
-        -- Check if the item will make the top 10 elements
+        -- Check if the item will make the top 5 elements
         if value > threshold or index <= 5 then
             local inserted = false
             values[player] = value
-            -- Find where in the top 10 to insert the element
+            -- Find where in the top 5 to insert the element
             for next_index, next_player in ipairs(sorted) do
                 if value > values[next_player] then
                     table.insert(sorted, next_index, player)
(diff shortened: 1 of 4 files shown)


modules/commands/search.lua
```diff
@@ -0,0 +1,167 @@
...
+:add_param('item', false, item_parse)
+:enable_auto_concat()
+:register(function(player, item)
+    local players = search_players(game.players, item)
+    if #players == 0 then return {'expcom-inv-search.results-none', item.name} end
+    local top_players = sort_players(players, recent_sort)
+    display_players(player, top_players, item)
+end)
+
+--- Return the index of the player, higher means they joined more recently
```
mark9064, line 136: Is this comment correct?

modules/commands/search.lua
```diff
@@ -0,0 +1,167 @@
...
+--- Custom sort function which only retains 5 greatest values
+local function sort_players(players, func)
+    local sorted = {}
+    local values = {}
+    local threshold = nil
+
+    -- Loop over all provided players
+    for index, player in ipairs(players) do
+        local value = func(player)
+        -- Check if the item will make the top 10 elements
```
mark9064, line 57: I might be misunderstanding this, but is this tracking 5 or 10 elements?

mark9064 (approved):
Just a couple things. LGTM

modules/commands/search.lua
Cooldude2606: Originally this sort selected 10 elements, this was changed to 5 after I realised that 10 was too many to display in chat.

PR #207, opened 2021-04-07
Report and warning count on embeds

Cooldude2606:
Fixed a bug with all embed fields being inline. Also added a player's warning and report count to the respective embeds.


mark9064 (approved):
:+1:

PR #208, opened 2021-04-15
Refactor of spawn area, now includes water.

Cooldude2606:
The config for spawn area has been refactored to give more granular control. The comments within the config have be unifed to a consistent format.
The spawn area will now include water and tile replacement is smarter only filling water tiles rather than the whole area.
Some settings in the existing config were ignored, these cases are now handled correctly.


config/spawn_area.lua
```diff
@@ -2,156 +2,179 @@
...
     },
-    afk_belts = { --- @setting afk_belts setting relating to afk belts round the spawn point
-        enabled=true, --- @setting enabled weather afk belts will be generated
-        locations={ --- @setting locations top left connors of any afk belt loops to be added
-            {x=-5,y=-5},
-            {x=5, y=-5},
-            {x=-5,y=5 },
-            {x=5, y=5 }
+    afk_belts = { --- @setting afk_belts Settings relating to adding afk belts to spawn
+        enabled = true, --- @setting enabled Weather afk belts will be added to spawn
```
mark9064, line 23:
```suggestion
        enabled = true, --- @setting enabled Whether afk belts will be added to spawn
```

mark9064: Weather -> Whether applies in many places, I won't flag them all you can just find and replace them

config/spawn_area.lua
```diff
@@ -2,156 +2,179 @@
...
+        belt_type = 'transport-belt', --- @setting belt_type The belt to used used as afk belts
```
mark9064, line 24: First `used` should be `be`?

modules/addons/spawn-area.lua
```diff
@@ -35,113 +42,133 @@ local function protect_entity(entity, set_force)
...
+        local x2 = (x+0.5)^2
+        for y = -fr, fr do -- loop over y
+            local y2 = (y+0.5)^2
+            local dst = x2+y2
+            local pos = {x=position.x+x, y=position.y+y}
+            if dst < dr2 then
+                -- If it is inside the decon radius always set the tile
+                table.insert(tiles_to_make, {name=decon_tile, position=pos})
+            elseif dst < fr2 and surface.get_tile(pos).collides_with('player-layer') then
+                -- If it is inside the fill radius only set the tile if its water
```
mark9064, line 139: Should be `it's` or `it is`

modules/addons/spawn-area.lua
```diff
@@ -35,113 +42,133 @@ local function protect_entity(entity, set_force)
...
--- generates an area with no water and removes entities in the decon area
-local function spawn_base(surface, position)
-    local dr = config.corrections.deconstruction_radius
-    local dr2 = dr^2
-    local dtile = config.corrections.deconstruction_tile
-    local pr = config.corrections.pattern_radius
-    local pr2 = pr^2
-    local ptile = surface.get_tile(position).name
-    if ptile == 'deepwater' or ptile == 'water' then ptile = 'grass-1' end
+-- Generates extra tiles in a set pattern set out in the config
```
mark9064, line 81: Minor style suggestion: would `set out` be better expressed as `as defined`?
Applies to all instances of set out

mark9064 (approved):
Looks good other than a few minor things. The general improvement in docs quality is great to see

PR #210, opened 2021-04-15
Admin map markers

Cooldude2606:
Adds map markers which can only be edited or removed by game admins. These markers are placed after entering admin marker mode with a command.


locale/en/commands.cfg
```diff
@@ -82,4 +82,11 @@ offline=You cannot connect as the server is currently offline: __1__
 none-matching=No servers were found with that name, if you used an address please append true to the end of your command.
 
 [expcom-lastlocation]
-response=Last location of __1__ was [gps=__2__,__3__]
\ No newline at end of file
+response=Last location of __1__ was [gps=__2__,__3__]
+
+[expcom-admin-marker]
+exit=You have left admin marker mode, all new makers will not be protected.
```
mark9064, line 88:
```suggestion
exit=You have left admin marker mode, all new markers will not be protected.
```
Same for line below

modules/commands/admin-markers.lua
```diff
@@ -0,0 +1,88 @@
...
+end)
+
+--- Listen for tags being removed or edited, maintain tags edited by non admins
+local function maintain_tag(event)
+    local tag = event.tag
+    if not event.player_index then return end
+    if not markers[tag.force.name..tag.tag_number] then return end
+    local player = game.get_player(event.player_index)
+    if player.admin then
+        -- Player is added, tell them it was an admin marker
```
mark9064, line 63: Should `added` be `admin` here? I'm not quite sure what it's trying to express

mark9064 (approved):
Just a couple bits :smile:

PR #211, opened 2021-04-16
Spectate and Follow

Cooldude2606:
Adds commands to toggle spectator mode and a command to follow a player.


modules/control/spectate.lua
```diff
@@ -0,0 +1,168 @@
...
+    assert(player and player.valid, 'Invalid player given to follower')
+    if spectating[player.index] or not player.character then return false end
+    local character = player.character
+    player.set_controller{ type = defines.controllers.spectator }
+    player.associate_character(character)
+    spectating[player.index] = character
+    return true
+end
+
+--- Return a player from spectator back to their character, if the character was killed then respawn the character
```
mark9064, line 44:
```suggestion
--- Return a player from spectator back to their character, if their character was killed then respawn the character
```

modules/control/spectate.lua
```diff
@@ -0,0 +1,168 @@
...
+}, function(tbl)
+    following = tbl.following
+    spectating = tbl.spectating
+end)
+
+----- Public Functions -----
+
+--- Test if a player is in spectator mode
+-- @tparam LuaPlayer player The player to test the controller type of
+-- @treturn boolean True if the player is in spectator mode
```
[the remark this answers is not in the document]
mark9064: Applies module wide

modules/control/spectate.lua
```diff
@@ -0,0 +1,168 @@
...
+    label.location = {0, res.height-150}
+    label.style.width = res.width
+    label.style.horizontal_align = 'center'
+    player.opened = label
+
+    return label
+end)
+:on_click(Public.stop_follow)
+:on_close(function(player)
+    -- Dont call set_controller during on_close as it invalidates the controller
```
mark9064, line 125:
```suggestion
    -- Don't call set_controller during on_close as it invalidates the controller
```
or `Do not` if you're afraid of apostrophes :P

modules/control/spectate.lua
```diff
@@ -0,0 +1,168 @@
...
+    assert(entity and entity.valid, 'Invalid entity given to follower')
+    local spectate = Public.start_spectate(player)
+
+    player.close_map()
+    follow_label(player.gui.screen, entity)
+    player.teleport(entity.position, entity.surface)
+    following[player.index] = { player, entity, entity.position, spectate }
+end
+
+--- Stops a player from following their current entity, can also return a player from spectate
```
mark9064, line 82: I'm not a fan of how this is worded, it almost implies that the player is prevented from following rather than them choosing to no longer follow

mark9064: Same below

mark9064 (approved):
Just a few things, looks good :+1:

modules/control/spectate.lua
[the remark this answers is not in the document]
Cooldude2606: Although it is redundant I have decided to use `Returns true when` as this creates a correct sentence.

modules/control/spectate.lua
```diff
@@ -0,0 +1,168 @@
...
+
+--- Test if a player is in spectator mode
+-- @tparam LuaPlayer player The player to test the controller type of
+-- @treturn boolean True if the player is in spectator mode
+function Public.is_spectating(player)
+    assert(player and player.valid, 'Invalid player given to follower')
+    return player.controller_type == defines.controllers.spectator
+end
+
+--- Puts a player into spectator while maintaining an association to their character
```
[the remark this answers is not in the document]
Cooldude2606: Although `spectator` is the name on the controler, I have decided to use `spectator mode` as their is additional logic implemented by this module.

modules/control/spectate.lua
Cooldude2606: Corrected to `if their character was killed then respawn them`

modules/control/spectate.lua
Cooldude2606: > or Do not if you're afraid of apostrophes :P

Absolutely not, they are far better than speach marks. :)

modules/control/spectate.lua
Cooldude2606: Changed to `Returns camera control to the player, will return a player to their character if start_follow placed them into spectator mode` and `Returns camera control to all players` as the function stops updating the position of their player/camera when called.

Issue #209, opened 2021-04-15
Protected Entities


Cooldude2606:
**Feature overview**
Name: Protected entities
TL;DR: Certain entities, such as water pumps, become protected when placed and can only be removed by the owner and regulars. 

**Details**
Within the config file there will be a set of entities that once placed will become protected and can not be removed by guests unless they palced it. This can include water pumps, steam engines, train stops etc. If a player attempts to mine one of these entities then a warning is shown in chat for online mods to see. Mods will also have a command which can be used to set a single entity as protected, and again to remove the protection.

**Possible methods**
Have a global array of protected entitites that is searched when a player with guest rank mines an entitiy. If they are not the owner then the entity is replaced and a message is displayed to mods.

**Additional context**
Maybe it would be better to have the warnings as alerts next to the toolbar. It should also be considered if the entity should be replaced by the script, as doing so would make greifers aware of the system and find another less detectable way to greif.

Cooldude2606:
Feature has been implemented in #213 and merged into dev branch. Closing issue.

PR #213, opened 2021-04-24
Entity protection

Cooldude2606:
When protected entities are removed by guests a message will be sent to discord and in chat. Protected entities are selected either indivdually or in an area.
Both methods use the new selection control module which impliments the selection planner. While the selection planner is in use all currelty protected areas and entities can be viewed.
Implements #209


modules/commands/protection.lua
```diff
@@ -0,0 +1,216 @@
...
+--- Remove a render object for a player
+local function remove_render(player, key)
+    local render = renders[player.index][key]
+    if render and rendering.is_valid(render) then rendering.destroy(render) end
+    renders[player.index][key] = nil
+end
+
+--- Toggles entity protection selection
+-- @command protect-entity
+Commands.new_command('protect-entity', 'Toggles entity protection selection, hold shift to remove protection')
```
mark9064, line 100: The command details are duplicated in locale - I'm guessing hardcoding them is the only way to do it

modules/commands/protection.lua
```diff
@@ -0,0 +1,216 @@
...
+        end
+    end
+    -- Show protected areas
+    local areas = EntityProtection.get_areas(surface)
+    for _, area in pairs(areas) do
+        show_protected_area(player, surface, area)
+    end
+end)
+
+--- When selection ends show hide protected entities on screen and protected areas
```
mark9064, line 203: `show hide` should be `hide`?

modules/control/protection.lua
```diff
@@ -0,0 +1,206 @@
...
+local EntityProtection = {
+    protected_entity_names = table.deep_copy(config.always_protected_names),
+    protected_entity_types = table.deep_copy(config.always_protected_types),
+    events = {
+        --- When a player mines a protected entity
+        -- @event on_player_mined_protected
+        -- @tparam number player_index the player index of the player who got mined the entity
+        -- @tparam LuaEntity entity the entity which was mined
+        on_player_mined_protected = script.generate_event_name(),
+        --- When a player mines a many protected entities
```
mark9064, line 19: Should `a many` be `many`? I'm not sure what it's saying

modules/control/protection.lua
```diff
@@ -0,0 +1,206 @@
...
+        -- @event on_repeat_violation
+        -- @tparam number player_index the player index of the player who got mined the entities
+        -- @tparam LuaEntity entity the last entity which was mined
+        on_repeat_violation = script.generate_event_name(),
+    }
+}
+
+-- convert config tables into lookup tables
```
mark9064, line 27: You normally capitalise docstrings right?

mark9064: Same for one below

modules/control/protection.lua
```diff
@@ -0,0 +1,206 @@
...
+local function get_area_key(area)
+    return string.format('%i,%i', math.floor(area.left_top.x), math.floor(area.left_top.y))
+end
+
+--- Check if an entity is always protected
+local function check_always_protected(entity)
+    return config.always_protected_names[entity.name] or config.always_protected_types[entity.type] or false
+end
+
+--- Check if an entity skips repeated
```
mark9064, line 77: Hmmm, I don't think this docstring really explains what's going on here. Is it checking whether an entity is exempt from repeated removal protection?

modules/control/selection.lua
```diff
@@ -0,0 +1,171 @@
...
+    - Controls players who have a selection planner, mostly event handlers
+    @control Selection
+    @alias Selection
+]]
+
+local Event = require 'utils.event' --- @dep utils.event
+local Global = require 'utils.global' --- @dep utils.global
+local Selection = {
+    events = {
+        --- When a player enterers selection mode
```
mark9064, line 11:
```suggestion
        --- When a player enters selection mode
```

modules/control/selection.lua
```diff
@@ -0,0 +1,171 @@
...
+end)
+--- Stop selection after an event such as death or leaving the game
+local function stop_after_event(event)
+    local player = game.get_player(event.player_index)
+    Selection.stop(player)
+end
+
+Event.add(defines.events.on_pre_player_left_game, stop_after_event)
+Event.add(defines.events.on_pre_player_died, stop_after_event)
+--- Stop selection after a single use if the option was used
```
mark9064, line 161: I think `if the option was used` is vague - could you make this more specific

mark9064 (approved):
Few bits here and there

config/protection.lua
```diff
@@ -0,0 +1,19 @@
...
+    always_protected_names = { --- @setting always_protected_names Names of entities which are always protected
+
+    },
+    always_protected_types = { --- @setting always_protected_types Types of entities which are always protected
+        'boiler', 'generator', 'offshore-pump', 'power-switch', 'reactor', 'rocket-silo'
+    },
+    skip_repeat_names = { --- @setting skip_repeat_names Names of entities which always trigger protection repeated
+
+    },
+    skip_repeat_types = { --- @setting skip_repeat_types Types of entities which trigger protection repeated
```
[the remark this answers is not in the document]
Cooldude2606: Changed endding to `always trigger repeated removal protection` to be in line with your next comment.

modules/commands/protection.lua
Cooldude2606: Hardcoding as strings is used to allow the search function of chelp to work.
When the commands module was made there was no way to access the locale string contents.
Even now it is only possible in an async manor, so although possible, it is not worth switching to locale strings.

modules/commands/protection.lua
Cooldude2606: Also removed `on screen` as this is no longer true.

modules/control/protection.lua
Cooldude2606: Changed to `When a player repeatedly mines protected entities` as this is even more clear.

modules/control/protection.lua
Cooldude2606: Although these ones arent docstring, I still normally capitalise for comments, so fixed that.

modules/control/protection.lua
Cooldude2606: Initially I used `skip` as it skiped the check for the repeat counter. It is much clear to use `always_trigger` and so all occurrences in the module and in the config have been changed to this.

modules/control/selection.lua
Cooldude2606: Chagned to refer to the `single_use` argument of `Selection.start`

PR #217, opened 2021-08-14
Fix auth in GUI not working.

oof2win2:


Cooldude2606:
Updating the top flow would need to be done inside the event emit as there are 2 other places where this would be needed.
I don't want Roles or Gui to depend on each other, so although this would be very small having its own file would be preferable.
Best solution might be to use `_C.opt_require` inside `gui/_require.lua` to check if the roles module exists then connect to it.

oof2win2:
I believe that a494391  should fix that

Cooldude2606 (approved):
All looks good to me.

Cooldude2606:
I will be treating this like a patch, so I am changing the base branch to merge into a patch.

PR #220, opened 2021-09-19
Fix tree decon toggle not working

oof2win2:
Actually check whether a user has the tree decon enabled/disabled before adding the entity to the decon queue.


Cooldude2606 (approved):
Looks good to me.

PR #223, opened 2021-09-26
Fix warp list crashing if no warp found

oof2win2:


Cooldude2606 (changes requested):
The proposed solution can result in an inconsistent state as `PlayerInRange` and `add_warp_element` will not be updated until a warp is found on your surface.
This results in an invalid state if `PlayerInRange` was true or `add_warp_element` was disabled before moving to the new surface.

oof2win2:
I think that I resolved this, assuming that the GUI should look like this when you cannot warp:

modules/gui/warp-list.lua
```diff
@@ -806,15 +808,14 @@ Event.on_nth_tick(math.floor(60/config.update_smoothing), function()
             local add_warp_element = frame.container.header.alignment[add_new_warp.name]
             local old_closest_warp_name = add_warp_element.tooltip[2] or closest_warp.name
             local was_able_to_make_warp = add_warp_element.enabled
-            local can_make_warp = closest_distance > mr2
+            local can_make_warp = closest_distance and closest_distance > mr2 or false
```
Cooldude2606, line 811: The default value should be true, as if no warp is in range then the player is able to make a warp.

modules/gui/warp-list.lua
```diff
@@ -765,7 +765,7 @@ Event.on_nth_tick(math.floor(60/config.update_smoothing), function()
 
         -- Check if the force has any warps
         local closest_warp
-        local closest_distance
+        local closest_distance = MAX_INT32_POS
```
Cooldude2606, line 768: `MAX_INT32_POS` is undefined and is unneeded as all other code changes work when `closest_distance` is nil.

modules/gui/warp-list.lua
```diff
@@ -806,15 +808,14 @@ Event.on_nth_tick(math.floor(60/config.update_smoothing), function()
             local add_warp_element = frame.container.header.alignment[add_new_warp.name]
             local old_closest_warp_name = add_warp_element.tooltip[2] or closest_warp.name
             local was_able_to_make_warp = add_warp_element.enabled
-            local can_make_warp = closest_distance > mr2
+            local can_make_warp = closest_distance and closest_distance > mr2 or false
             if can_make_warp and not was_able_to_make_warp then
                 add_warp_element.enabled = true
                 add_warp_element.tooltip = {'warp-list.add-tooltip'}
-            elseif not can_make_warp and was_able_to_make_warp or old_closest_warp_name ~= closest_warp.name then
+            elseif not can_make_warp and was_able_to_make_warp or closest_warp and old_closest_warp_name ~= closest_warp.name then
```
Cooldude2606, line 815: Given that when `closest_distance` is nil, `can_make_warp` will be true, the final condition of this statement will never be evaluated so `closest_warp and` can be removed.

Cooldude2606 (changes requested):

modules/gui/warp-list.lua
oof2win2: Oh, i thought it meant that the user is able to warp.

modules/gui/warp-list.lua
oof2win2: This doesn't work, because if `can_make_warp` is true and `was_able_to_make_warp` is also true, the first condition does not execute. The issue I found is that `was_able_to_make_warp` does not seem to be toggling itself

Cooldude2606: You are right on `closest_warp and old_closest_warp_name ~= closest_warp.name` being evaluated, I was wrong in thinking it would not as I mistook the `or` for an `and`.
`was_able_to_make_warp` is set based on the enabled state of the gui element. Therefore the toggle is done when `add_warp_element.enabled` is set in these if statements. However, I do not see what issue you are refering to... if `closest_warp` is nil then `can_make_warp` is true; then if `was_able_to_make_warp` is false then the button becomes enabled and `was_able_to_make_warp` will be true the next time, otherwise if its true then the second statement does execute and the button remains enabled.

Cooldude2606: Ah I see the issue is because you have done `or true` so even when `can_make_warp` should be false it is becoming true. This would be more clear to see if you gave your self the make warp permission, you currently do not if your last screenshot is to go by.

oof2win2: Pretty sure that I fixed it with 5d65144

PR #224, opened 2021-12-29
Basic Action Log

oof2win2:
Adds a basic action log (not just decon) that logs deconning areas, built entities and mined entities. Bypassed with the 'Member' role by default to prevent spam.


Cooldude2606 (changes requested):
All code looks correct, however I would like a config file added similar to discord alerts. That is a dictionary of booleans which decide which events are logged.

oof2win2:
Implemented.

Cooldude2606 (approved):
lgtm. This module can be easilly extended in the future to add more events should they be wanted.

PR #227, opened 2022-05-14
Fix transfer of items after death and ban

oof2win2:
Fixed an issue where items would be transferred only by name and "copied", which would cause issues with items with metadata, for example contents of power armor.


Cooldude2606 (changes requested):
All code looks good; however, their should be a note in the doc comment about how this function is different to `move_items`

oof2win2:
Resolved

Cooldude2606 (approved):
LGTM.

PR #222, opened 2021-09-25
Graftorio addition

oof2win2:
Add graftorio for Grafana integration of some sweet production graphs etc.
Sourced from [TheVirtualCrew/graftorio](https://github.com/TheVirtualCrew/graftorio) (modified for scenario usage and not a mod) with big thanks to Tim Vroom.

PR #228, opened 2022-05-14
Add graftorio

oof2win2:
Partially add some bits of [graftorio](https://github.com/TheVirtualCrew/graftorio). Data can be collected with the `/collectdata` command and is only sent to RCON currently.
See [this](https://github.com/oof2win2/AwF-Bot/blob/master/src/base/GrafanaHandler.ts) for implementation of handling the data in JS.


modules/graftorio/require.lua
```diff
@@ -0,0 +1,23 @@
...
+Commands.new_command("collectdata", "Collect data for RCON usage")
+		:add_param("location", true)
+		:register(function()
+			-- this must be first as it overwrites the stats
+			-- also makes the .other table for all forces
+			statics.collect_statics()
+			general.collect_other()
+			if config.modules.forcestats then
+				forcestats.collect_production()
+			end
```
Cooldude2606, line 20: `collect_loginet` is never called which is part of the `forcestats` module.
`general` has a config which is never tested for.
```suggestion
			statics.collect_statics()
			if config.modules.general then
			  general.collect_other()
			end
			if config.modules.forcestats then
				forcestats.collect_production()
				forcestats.collect_loginet()
			end
```

modules/graftorio/require.lua
```diff
@@ -0,0 +1,23 @@
+local Commands = require("expcore.commands")
+local config = require("config.graftorio")
+
+local general = require("modules.graftorio.general")
```
Cooldude2606, line 4: `general` needs a similar test which forcestats has.
```suggestion
local general; if config.modules.general then
  general = require("modules.graftorio.general")
end
```

Cooldude2606 (changes requested):
Two small corrections required. I will also have mark review this PR due to his familiarity with grafana.

Cooldude2606 (approved):
Changes have been addressed. LGTM.

config/expcore/roles.lua
```diff
@@ -52,6 +52,7 @@ Roles.new_role('Administrator','Admin')
     'gui/warp-list/bypass-cooldown',
     'gui/warp-list/bypass-proximity',
     'command/connect-all',
+		'command/collectdata'
```
mark9064, line 55: Might just be github but there seems to be a whitespace issue here?

config/graftorio.lua
```diff
@@ -0,0 +1,7 @@
+return {
+	modules = {
+		["forcestats"] = true,
+		["logistorage"] = false,
```
mark9064, line 4: We aren't shipping the `logistorage` module, so we can drop this for now?

mark9064 (commented):
Merged just as I started the review, oh well! LGTM overall anyway.
One potentially important thing: we probably need to acknowledge the graftorio license if we are using their code here directly.

config/graftorio.lua
oof2win2: logistorage itself is a part of forcestats collected with robot metrics. might be worth splitting it apart in the future

config/expcore/roles.lua
oof2win2: you are right there.

oof2win2:
We don't need to acknowledge the graftorio licence as this is basically a rewrite of it that is suitable for the scenario and doesn't have features that I deemed unnecessary, such as translation of prototype names. Graftorio itself by TheVirtualCrew also outputs into a file. It would have been more appropriate to add the licence if #222 was merged since that was mainly their code, but it wasn't merged and this is a rewrite.

PR #229, opened 2022-05-14
Implement nuke protect, fix move_items

oof2win2:
Changed _C.move_items* to be _C.copy_items_stack, which copies items, and _C.move_items_stack, which moves items.

Added an addon called `nukeprotect`, which prevents players without a flag from having specific items in their inventories.

### DIFF
--- a/config/expcore/roles.lua
+++ b/config/expcore/roles.lua
@@ -223,7 +223,8 @@ Roles.new_role('Regular','Reg')
     'command/go-to-spawn',
     'command/me',
     'standard-decon',
-    'bypass-entity-protection'
+    'bypass-entity-protection',
+		'bypass-nukeprotect'
 }
 :set_auto_assign_condition(function(player)
     if player.online_time >= hours3 then
--- a/config/expcore/roles.lua
+++ b/config/expcore/roles.lua
@@ -291,7 +291,7 @@ Roles.define_role_order{
 }
 
 Roles.override_player_roles{
-    ["oof2win2"]={"Senior Administrator","Moderator","Senior Backer","Supporter"},
+    ["Cooldude2606"]={"Senior Administrator","Moderator","Senior Backer","Supporter"},
     ["arty714"]={"Senior Administrator","Senior Backer","Supporter"},
     ["Drahc_pro"]={"Administrator","Moderator","Veteran","Member"},
     ["mark9064"]={"Administrator","Moderator","Member"},
--- a/config/expcore/roles.lua
+++ b/config/expcore/roles.lua
@@ -291,7 +291,7 @@ Roles.define_role_order{
 }
 
 Roles.override_player_roles{
-    ["Cooldude2606"]={"Senior Administrator","Moderator","Senior Backer","Supporter"},
+    ["oof2win2"]={"Senior Administrator","Moderator","Senior Backer","Supporter"},
     ["arty714"]={"Senior Administrator","Senior Backer","Supporter"},
     ["Drahc_pro"]={"Administrator","Moderator","Veteran","Member"},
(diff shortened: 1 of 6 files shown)


config/nukeprotect.lua
```diff
@@ -0,0 +1,11 @@
+return {
+	ammo = {
+		["atomic-bomb"] = true
+	}, -- @setting ammo The items to not allow in the player's ammo inventory
+	armor = {}, -- @setting armor The items to not allow in the player's armor inventory
+	gun = {}, -- @setting gun The items to not allow in the player's gun inventory
+	main = {
+		["atomic-bomb"] = true
+	}, -- @setting main The items to not allow in the player's main inventory
+	ignore_permisison = "bypass-nukeprotect" -- @setting ignore_permisison The permission that nukeprotect will ignore
```
Cooldude2606, line 10: I would include an option for `ignore_admins` which is true by default, this is to allow the module to be used outside of the Roles module.

modules/addons/nukeprotect.lua
```diff
@@ -0,0 +1,75 @@
...
+
+	local inv = player.get_inventory(defines.inventory.character_ammo)
+
+	for i = 1, #inv do
+		local item = inv[i]
+		if item.valid and item.valid_for_read and config.ammo[item.name] then
+			player.print({ "nukeprotect.ammo", { "item-name." .. item.name } })
+			move_items_stack({ item })
+		end
+	end
```
Cooldude2606, line 23: This can be moved into its own function to prevent repetition.

modules/addons/nukeprotect.lua
```diff
@@ -0,0 +1,75 @@
+--- Disable new players from having certain items in their inventory, most commonly nukes
+-- @addon Nukeprotect
+
+local Event = require 'utils.event' --- @dep utils.event
+local Roles = require 'expcore.roles' --- @dep expcore.roles
+local config = require 'config.nukeprotect' --- @dep config.nukeprotect
+local move_items_stack = _C.move_items_stack --- @dep expcore.common
+
+Event.add(defines.events.on_player_ammo_inventory_changed, function(event)
```
Cooldude2606, line 9: Because the config is static, you can avoid adding handlers if the arrays are empty.
eg `if #config.ammo > 0 then --[[ register ammo handler --]] end`

Cooldude2606 (changes requested):

oof2win2:
I added the new config, though I'm not sure if this is a good way to do it so that it is "modular".

config/expcore/roles.lua
```diff
@@ -291,7 +291,7 @@ Roles.define_role_order{
 }
 
 Roles.override_player_roles{
-    ["Cooldude2606"]={"Senior Administrator","Moderator","Senior Backer","Supporter"},
+    ["oof2win2"]={"Senior Administrator","Moderator","Senior Backer","Supporter"},
```
Cooldude2606, line 294: Accidental commit.

Cooldude2606 (changes requested):
Firstly, I dont see any reason to use `tostring` on the keys.
Secondally, if you want to make it modular then you should include the event name in the config:
```
{
  {
    inventory = defines.inventory.character_ammo,
    event = defines.events.on_player_ammo_inventory_changed,
    items = {
      ["atimic-bomb"] = true
    }
  }
}
```
This will alow you to use a for loop within the module code.
```lua
for _, area in ipairs(config) do
  if #area.items > 0 then
    Event.add(area.event, ...)
  end
end
```

oof2win2:
Implemented the desired modularity of the code

Cooldude2606 (approved):
LGTM

PR #231, opened 2022-09-14
Add ban and unban logging for FAGC clientside bot support

oof2win2:
https://github.com/FactorioAntigrief/FactorioAntigrief/issues/124

### DIFF
--- a/modules/addons/fagc.lua
+++ b/modules/addons/fagc.lua
@@ -5,15 +5,15 @@ local Event = require 'utils.event' --- @dep utils.event
 
 -- Clear the file on startup to minimize its size
 Event.on_init(function(e)
-	game.write_file("fagc-actions.txt", "", true, 0)
+	game.write_file("fagc-actions.txt", "", false, 0)
 end)
 
 Event.add(defines.events.on_player_banned, function(e)
-	local text = "ban;\"" .. e.player_name .. "\";\"" .. (e.by_player or "") .. "\";\"" .. (e.reason or "") .. "\""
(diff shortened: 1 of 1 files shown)


modules/addons/fagc.lua
```diff
@@ -0,0 +1,19 @@
+--- Allows the FAGC clientside bot to receive information about bans and unbans and propagate that information to other servers
+-- @addon FAGC
+
+local Event = require 'utils.event' --- @dep utils.event
+
+-- Clear the file on startup to minimize its size
+Event.on_init(function(e)
+	game.write_file("fagc-actions.txt", "", true, 0)
```
Cooldude2606, line 8: This needs to be false in order to overwrite the file.

modules/addons/fagc.lua
```diff
@@ -0,0 +1,19 @@
...
+end)
+
+Event.add(defines.events.on_player_banned, function(e)
+	local text = "ban;\"" .. e.player_name .. "\";\"" .. (e.by_player or "") .. "\";\"" .. (e.reason or "") .. "\""
```
Cooldude2606, line 12: Why use both quotes and semi-colons where neither are possible in usernames.
It may even be better to keep this csv compliant, meaning use `,` to seperate each part and use `"` around the reason.

Cooldude2606 (changes requested):

modules/addons/fagc.lua
oof2win2: Resolved

Cooldude2606 (approved):

PR #232, opened 2022-10-20
Update deconlog.lua

Windsinger1985:
Add for regular rocket and explosive rocket + adding comma between timestamp and playername.


Cooldude2606 (changes requested):
All code looks good to me; however, you should add the default values into `config/deconlog.lua`

Windsinger1985:
> All code looks good to me; however, you should add the default values into `config/deconlog.lua`

Took me a moment to realise what i missed here, sorry for that, added it in the PR as well.

Cooldude2606 (approved):
LGTM.

PR #234, opened 2022-10-23
Patch/6.2.1

Cooldude2606:
Fixes:
- Report reasons being reset to "No Reason Given"
- Readme now uses refernces for links
- Removed ending slashes for URLs
- Fixed broken link in readme
- Topflow from ExpCore/Gui correctly updates when the authenticator function returns false.


oof2win2 (approved):
LGTM.

PR #235, opened 2022-11-12
fix: set roles after jailing

oof2win2:


Cooldude2606 (changes requested):
The logic is almost there; however, I have since discovered aditional edge cases which will not be covered. 
Because of this I suggest moving the logic into `role:add_player` and `role:remove_player`.

You also need to correct the `= 1` and `= -1` to be `+=` and `-=` (lua equivalent) and use `< 0` and `> 0` rather than `== -1` and `== 1`.

Cooldude2606:
@oof2win2 I have implimented the changes I requested in order to get this merged for the upcoming release.
Could you please validate that the bug you reported is now fixed.

oof2win2:
Sorry I didn't get back to this. Currently can't, I'd be able to get back to you on the 25th or 26th. It looks fine by me though if that's enough to merge it

oof2win2:
Looks good to me 👍

Cooldude2606 (approved):
Self sign-off of own changes after validation from OP.

Cooldude2606:
Bypassing review requirement: New changes are from rebase and so do not include any new code.

PR #237, opened 2023-02-13
Feature Update

PHIDIAS0303:
- [x] **1 Sever UPS Display**
Slightly adjust the display a bit lower so it don't block the clock
Additional game speed percentage display

- [X] **2 Deconstruction Planner Chat Display**
Show a message to notify other.

- [X] **3 Advanced Start**
Allow an even faster start with the modular armor building kit.

- [X] **4 Report Jail**
Now it looks up if an active moderator is in-game.
Otherwise, then uses logic to check player data once.

- [X] **5 Protection Jail**
Now it looks up if an active moderator is in-game.

- [x] **6 Spawn Area**
Optional edit to allow customized generation.
This slightly boosts the nearby patch so it is faster to start.

- [X] **7 Teleport With Vehicle**
teleport and spawn should be the same as warp
#236 

- [X] **8 Vlayer**
Optional virtual layer module to allow sending solar panel and battery inside.
This could save a lot of space.

- [X] **9 Personal Logistic Slot**
The optional setting for the personal logistic slot. only for members and above.

- [X] **10 Auto Miner Deconstruction**
Optional, an attempt for automatic miner deconstruction.
- [X] **11 Research Notification**
display the research completed message

- [X] **12 Auto Research**
Auto research infinite research

- [X] **13 Discord Alert of Protected Entities**
The chat bridge is missing a message display so the removed machine is nowhere to be found without a location written on the message.

- [X] **14 Bot Queue**
Set the construction bot queue.

- [X] **15 Game Speed**
Set the game speed.

- [X] **16 Pollution**
Modify game pollution.

- [X] **17 Manual train**
one click automatic

- [X] **18 Friendly Fire**
for the modded server if needed

- [X] **19 Personal Battery Recharge**
using the vlayer energy

- [X] **20 Lawnmower**
Clean up biter corpse and nuclear hole

some other features will be update in the next version.

PR #241, opened 2023-04-05
Energy surface working draft

bbassie:
Stolen idea from the pull request from @PHIDIAS0303 but then with less "magic". Would need to be developed further if there is interest in this method:
Dropped 25 solar in the chest and 25 accumulators.
image
energy-surface:
image
Substation connects back to one of the spawn power poles. No logic implemented for reconnection for when this connection is broken!

Most code "generated" with copilot so be nice 😊


Cooldude2606:
The PR this idea was taken from (#237) has now been merged with dev. So if you chose to pick this up again it will need be rebased on the existing VLayer feature.

PR #245, opened 2023-09-06
Disable nukes completely

oof2win2:
This PR disables nukes completely by disabling the research for them (configgable), as they don't have much of real use besides griefers using them.

### DIFF
--- a/config/nukeprotect.lua
+++ b/config/nukeprotect.lua
@@ -26,5 +26,9 @@ return {
 		},
 	},
 	ignore_permisison = "bypass-nukeprotect", -- @setting ignore_permisison The permission that nukeprotect will ignore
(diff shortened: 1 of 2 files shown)


Cooldude2606 (commented):
I think the default should be "enabled" because in the late game people use spidertrons with nukes. However, this feature is still useful for server that do not reach that stage of the game so this can still be accepted.

oof2win2:
Fixed those changes. With experience from our community, most people just abuse nukes overall since the blast radius is too large for spidertrons or rocket launchers.

Cooldude2606 (approved):
I spoke with moderators within our community and the consensus was that the greatest benefit from disabling nukes is that they can no longer be used to create lag. However, artillary shells are a much cheeper, simpler, and more effective way to achieve the same result. Therefore, there is not alot to be gained from disabling them.

PR #248, opened 2023-09-22
addon module inserter gui

PHIDIAS0303:


modules/gui/module.lua
```diff
@@ -0,0 +1,267 @@
...
+local SelectionModuleArea = 'ModuleArea'
+
+--[[
+for some reason this dont work
+
+local module_allowed = {}
+
+for _, r in pairs(game.item_prototypes['productivity-module'].limitations) do
+    module_allowed[r] = true
+end
```
Cooldude2606, line 18: This does not work because `game` is not valild during the control stage of factorio. You would need to this call during an event for it to be valid.

config/module.lua
```diff
@@ -0,0 +1,20 @@
+return {
+    -- type of machine to handle together
+    module_row = 4,
```
Cooldude2606, line 3: Could you rename this `default_row_count` because it was not obvious what it was doing until i read the code.

Cooldude2606 (changes requested):
Some minor changes, only one breaking change with how you handle the gui element events.

config/module.lua
Cooldude2606: This has been updated to `default_module_row` please use the word "count" in it to describe what it does.

Cooldude2606 (approved):
All changes were made correctly, no conflicts after update to latest dev branch.

PR #249, opened 2023-09-23
cmd personal logistics for spidertron

PHIDIAS0303:


modules/data/personal-logistic.lua
```diff
@@ -63,12 +73,32 @@ local function pl(player, amount)
...
 
-Commands.new_command('personal-logistic', 'Set Personal Logistic (0 to cancel all)')
+Commands.new_command('personal-logistic', 'Set Personal Logistic (0 to cancel all) (Select spidertron to edit spidertron)')
 :add_param('amount', 'integer-range', 0, 10)
 :add_alias('pl')
 :register(function(player, amount)
     if player.force.technologies['logistic-robotics'].researched then
-        pl(player, amount / 10)
+        if player.selected ~= nil then
+            if player.selected.name ~= nil then
```
Cooldude2606, line 82: `name` is never nil. This check is not required.

modules/data/personal-logistic.lua
```diff
@@ -1,30 +1,40 @@
...
+            c = target.clear_personal_logistic_slot
+            s = target.set_personal_logistic_slot
 
-    for k, v in pairs(config.request) do
+            for k, v in pairs(config.request) do
+                c(config.start + v.key)
+                s(config.start + v.key, {name=k, min=0, max=0})
+            end
+
+            return
```
Cooldude2606, line 25: There is no benifit to include "clear" in this function. Either make it into its own function or place the for loop directly in the command body.
After you have done this there will be no difference between "player" (`p`) and "selected" (`s`) so you can remove type from the arguments and set `c` and `s` unconditionally.

Cooldude2606 (changes requested):
Small improvements can be made, but functionaly is all correct.

Cooldude2606 (approved):
All looks good. You can still have the second command if you want, my change request was that it should be in its own function.

PHIDIAS0303:
it is not really needed for the 2nd cmd i think

PR #251, opened 2023-09-25
CCTV module

PHIDIAS0303:


modules/gui/surveillance.lua
```diff
@@ -0,0 +1,171 @@
...
+        scroll_table_1.add{
+            type = 'drop-down',
+            name = 'cctv_display_' .. i .. 'e',
+            items = {'Enable', 'Disable'},
+            selected_index = 1
+        }
+
+
+        local button_ds =
+        Gui.element{
```
Cooldude2606, line 46: Don't define new elements inside another element, because this will cause a new element to be defined for every player. Also `Gui.element` is not runtime safe but I forgot to include the runtime assertion which is why it did not error.

modules/gui/surveillance.lua
```diff
@@ -0,0 +1,171 @@
...
+    end
+end
+
+Event.add(defines.events.on_player_joined_game, function(_)
+    gui_update()
+end)
+
+Event.add(defines.events.on_player_left_game, function(_)
+    gui_update()
+end)
```
Cooldude2606, line 142: Rather than `, function(_) gui_update() end)` you can use `, gui_update)`

modules/gui/surveillance.lua
```diff
@@ -0,0 +1,171 @@
...
+
+Event.on_nth_tick(1, function()
```
Cooldude2606, line 144: Don't use `on_nth_tick(1,` use `on_tick` or `on_player_moved`

Cooldude2606 (changes requested):
Code is not runtime safe so must be changed.

PHIDIAS0303:
i have no idea what i am doing

modules/gui/surveillance.lua
```diff
@@ -0,0 +1,167 @@
...
+}
+
+local cctv_location =
+Gui.element{
+    type = 'button',
+    caption = 'set'
+}:style{
+    width = 40
+}:on_click(function(player, element, _)
+    element.parent.parent.cctv_display.position = player.position
```
Cooldude2606, line 53: Very simple correction here, this one is probably on me when I changed the GUI structure.
```suggestion
    element.parent.parent.parent.cctv_display.position = player.position
```

Cooldude2606 (commented):
Two small changes need, mostly around the size of buttons.
Could you make the cameras disabled by default, it feels odd to have both enabled on your player by default.
Once that is complete it can be merged.

modules/gui/surveillance.lua
```diff
@@ -0,0 +1,168 @@
...
+
+    camera.visible = false
+    camera.style.minimal_width = 480
+    camera.style.minimal_height = 360
+    return camera_set
+end)
+
+cctv_container =
+Gui.element(function(event_trigger, parent)
+    local container = Gui.container(parent, event_trigger, 480)
```
Cooldude2606, line 110: I didn't realise until testing, but we need a scroll pane here to stop the built in one from showing up.
400 seams like a good size because it fits one set of buttons and one camera.
You will also need to change `camera_set(container,` to `camera_set(scroll,` and change `gui_update` and `on_tick`
```lua
local scroll = container.add{ name="scroll", type="scroll-pane", direction="vertical" }
scroll.style.maximal_height = 400
```

modules/gui/surveillance.lua
```diff
@@ -108,14 +111,16 @@ end)
 cctv_container =
 Gui.element(function(event_trigger, parent)
     local container = Gui.container(parent, event_trigger, 480)
+    local scroll = container.add{name='scroll', type='scroll-pane', direction='vertical'}
+    scroll.style.maximal_height = 816
```
Cooldude2606, line 115: `maximal_height` must be `705` or less to prevent the external scroll bar.

Cooldude2606:
When a player leaves the game they are removed from the selection list. This results in the selected index returning `0` for "none selected". You need to include a check for this in on_tick to prevent error spam in the logs.
`Given player value (0) is too small, the data type allows values from 1 to 65536`

Cooldude2606 (approved):
LGTM. All changes accounted for. Gui could be improved, but it works.

PR #252, opened 2023-09-30
fix no recipe module insertion

PHIDIAS0303:


modules/gui/module.lua
```diff
@@ -123,15 +123,22 @@ end
 
 local function apply_module(player, area, machine, module)
     for _, entity in pairs(player.surface.find_entities_filtered{area=area, name=machine, force=player.force}) do
-        local m_current_recipe = entity.get_recipe()
+        if config.machine_craft[machine] ~= nil then
+            if config.machine_craft[machine] then
```
Cooldude2606, line 127: `if config.machine_craft[machine] ~= nil then` is not required.

Cooldude2606 (changes requested):
Only one change needed.

Cooldude2606 (approved):
LGTM.

PR #254, opened 2023-10-03
Vlayer Update

PHIDIAS0303:


Cooldude2606 (changes requested):
`vlayer` needs to be a module. This means you create an empty table at the top of your file and return it afterwards. You then add your functions to this table. See `tasks.lua` for an example. The most important part is you can not store functions in Global.

Cooldude2606 (dismissed):
I've made all the required changes I needed to, from my viewpoint it is now ready to merge.

Cooldude2606 (approved):

Issue #226, opened 2022-03-14
Other Player Data Display


PHIDIAS0303:
**Feature overview**
add a button in the player list like kill, ban.

**Possible methods**
a pop-up screen like info button to show player data info.
it would be great if ratio or what is included?
only show at that time data. no need live update.

PR #256, opened 2023-10-05
playerdata

PHIDIAS0303:


PHIDIAS0303:
#226

Cooldude2606 (changes requested):
It would be nice if changing the selected player would update the labels automatically. You can do this by moving the button on_click into a `update_labels` function which is used by both the on click and on selection change.

Also a big FYI, the player data stats are designed to be iterable. See the readme window for an example:
```lua
for _, name in pairs(PlayerData.Statistics.metadata.display_order) do
    local child = PlayerData.Statistics[name]
    local metadata = child.metadata
    local value = child:get(player_name)
    if value ~= nil or metadata.show_always then
        if metadata.stringify then
            value = metadata.stringify(value)
        else
            value = format_number(value or 0)
        end
        Gui.centered_label(statistics, 150, metadata.name or {'exp-statistics.'..name}, metadata.tooltip or {'exp-statistics.'..name..'-tooltip'})
        Gui.centered_label(statistics, 130, {'readme.data-format', value, metadata.unit or ''}, metadata.value_tooltip or {'exp-statistics.'..name..'-tooltip'})
    end
end
```

modules/gui/playerdata.lua
```diff
@@ -0,0 +1,755 @@
...
+    local player_list = {}
+
+    for _, player in pairs(game.players) do
+        table.insert(player_list, player.name)
+    end
+
+    pd_username_set(container, 'pd_st_1', player_list)
+    pd_data_set(container, 'pd_st_2')
+
+    local table = container['pd_st_2'].disp.table
```
Cooldude2606, line 730: This needs to be `pd_st_1` not `2`. 
`modules/gui/playerdata.lua:731: attempt to index field '?' (a nil value)`

modules/gui/playerdata.lua
```diff
@@ -0,0 +1,755 @@
...
+    table[pd_username_player.name].width = 224
```
Cooldude2606, line 731: This needs to be `.style.width`.
`LuaGuiElement doesn't contain key width.`

config/expcore/roles.lua
```diff
@@ -139,7 +139,8 @@ Roles.new_role('Board Member','Board')
     'command/repair',
     'command/spectate',
     'command/follow',
-    'command/personal-battery-recharge'
+    'command/personal-battery-recharge',
+    'gui/playerdata'
```
Cooldude2606, line 143: Can you also give the permission `gui/playerdata` to `Moderator`.

modules/gui/playerdata.lua
```diff
@@ -0,0 +1,755 @@
...
+end)
+
+local pd_username_update =
+Gui.element{
+    type = 'button',
+    caption = 'update'
+}:style{
+    width = 128
+}:on_click(function(_, element, _)
+    local player_name = game.players[element.parent[pd_username_player.name].selected_index]
```
Cooldude2606, line 702: You need to consider the case when `selected_index` is `0` because a player has left the game.

modules/gui/playerdata.lua
```diff
@@ -0,0 +1,755 @@
...
+:add_to_left_flow()
+
+Gui.left_toolbar_button('item/power-armor-mk2', 'Player Data GUI', pd_container, function(player)
+	return Roles.player_allowed(player, 'gui/playerdata')
+end)
+
+local function gui_player_list_update()
+    local player_list = {}
+
+    for _, player in pairs(game.players) do
```
Cooldude2606, line 744: This must be `game.connected_players` because player data is not valid for offline players.

Cooldude2606 (changes requested):

Cooldude2606 (commented):
Concept works, I have cleaned up alot of it tho. Could you please review it yourself to confirm all functions are the same as you intended.

PR #257, opened 2023-10-06
fixed non prod module insertion with recipe

PHIDIAS0303:


Cooldude2606 (changes requested):
The check you are doing would disallow you from inserting 1 prod and 1 speed module into a machine where prod is disallowed. The expected behaviour is for the item request to be 1 speed module.

Cooldude2606 (approved):
LGTM. Made final edit my self to speed up merge.

PR #258, opened 2023-10-13
1 decimal place on last location gps

PHIDIAS0303:
it doesnt seem there is a need to display full gps location


Cooldude2606 (approved):
Simple change. LGTM.

Issue #162, opened 2020-05-30
Decon trees and rocks if you drive into them.


Cooldude2606:
**Feature overview**
Name: Crash Decon
TL;DR: when you are in a car and you hit a tree or a rock then it mark it to be deconstructed, option to have this be turned off and on.

**Details**
Idea from: https://forums.factorio.com/viewtopic.php?f=33&t=85223

**Possible methods**
Use of on_entity_damaged and then test if the player is in a car and filter for trees and rocks. Then just mark them to be deconstructed. Using a command or gui to toggle would also be possible.

oof2win2:
Same thread has code that would work.
```lua
script.on_event(defines.events.on_entity_damaged, function(event)
	local entity = event.entity
	local force = event.force
	if entity.type == "tree" and event.damage_type.name == "impact" and force then
		entity.order_deconstruction(force)
	end
end)
```

PHIDIAS0303:
this should be done now

PR #259, opened 2023-10-13
deconstruction of tree and rocks when drive over them

PHIDIAS0303:
#162


Cooldude2606 (approved):
All looks good. Made an edit my self to allow the fast decon toggle to effect car collisions.

PR #261, opened 2023-10-24
artillery target remote

PHIDIAS0303:
maybe there are some better ways to calculate how to reduce duplicates, but I haven't thought of one.


Cooldude2606 (dismissed):
LGTM. The `location_break` check could be improved, but the edge cases are rare and hard to exploit, so its fine as is.

PR #264, opened 2023-12-29
CCTV location follow issue

PHIDIAS0303:


Cooldude2606 (approved):
I don't like the variable reuse, but it works.

PR #268, opened 2024-01-07
Feature: Toolbar Menu

Cooldude2606:
Adds a toolbar menu which allows the player to select the order of the toolbar buttons and left flow menus. It also allows buttons to be hidden from the top flow and interacted with directly within the menu. This all saves to the datastore and will sync across games.


Cooldude2606:
Had people live test this branch, any bugs found were fixed.

PR #270, opened 2024-01-17
miner - fluid and chest handle

PHIDIAS0303:


modules/addons/miner.lua
```diff
@@ -1,64 +1,209 @@
...
+
+    if not entity.prototype.selectable_in_game then
+        -- if it can select
+        return
+    end
+
+    if entity.has_flag('not-deconstructable') then
+        -- if it can deconstruct
+        return
+    end
```
Cooldude2606, line 52: Duplicate code, Pull this out to its own function so it can be shared between `miner_check` and `chest_check`

modules/addons/miner.lua
```diff
@@ -1,64 +1,209 @@
...
+
+    if target.type ~= 'logistic-container' and target.type ~= 'container' then
+        -- not a chest
+        return
+    end
+
+    local radius = 1 + entity.prototype.mining_drill_radius
+    local entities = entity.surface.find_entities_filtered{area={{entity.position.x - radius, entity.position.y - radius}, {entity.position.x + radius, entity.position.y + radius}}, type={'mining-drill', 'inserter'}}
```
Cooldude2606, line 60: Duplicate code, should be moved to its own function so it can be shared between `miner_check` and `chest_check`.
Also this checks should be done on the target not the mining drill passed to this function because those checks have already been done.

modules/addons/miner.lua
```diff
@@ -1,64 +1,209 @@
...
+        local entities = entity.surface.find_entities_filtered{area={{entity.position.x - radius, entity.position.y - radius}, {entity.position.x + radius, entity.position.y + radius}}, type={'mining-drill', 'pipe', 'pipe-to-ground'}}
+        local entities_t = entity.surface.find_entities_filtered{area={{entity.position.x - radius, entity.position.y - radius}, {entity.position.x + radius, entity.position.y + radius}}, ghost_type={'pipe', 'pipe-to-ground'}}
+        local c = #entities
 
-                if not entity.minable then
-                    -- if it is minable
-                    return
+        for k,v in pairs(entities_t) do
+            entities[c + k] = v
+        end
```
Cooldude2606, line 136: Use library method: `table.array_insert(entities, entities_t)`

modules/addons/miner.lua
```diff
@@ -1,64 +1,209 @@
...
+                for h=1, half do
+                    table.insert(pipe_build, {x=0, y=-h})
+                end
             end
         end
+
+        table.insert(miner_data.queue, {t=game.tick + 5, e=entity})
+
+        for p=1, #pipe_build do
+            entity.surface.create_entity{name='entity-ghost', position={x=entity.position.x + pipe_build[p].x, y=entity.position.y + pipe_build[p].y}, force=entity.force, inner_name='pipe', raise_built=true}
```
Cooldude2606, line 164: Would be better to do the addition in the loops above then use directy assignment here:
```lua
for _, pos in ipairs(pipe_build) do
    entity.surface.create_entity{name='entity-ghost', position=pos, force=entity.force, inner_name='pipe', raise_built=true}
end
```

modules/addons/miner.lua
```diff
@@ -1,64 +1,209 @@
...
+        end
+
+        if config.chest then
+            chest_check(entity)
+        end
```
Cooldude2606, line 169: Duplicate code, same for the `table.insert` above, instead allow for a drop through into the else statement. (Remove else)

modules/addons/miner.lua
```diff
@@ -1,64 +1,209 @@
...
+        if not q.e or not q.e.valid then
+            table.remove(miner_data.queue, k)
+            break
+
+        elseif event.tick >= q.t then
+            q.e.order_deconstruction(q.e.force)
+            table.remove(miner_data.queue, k)
+        end
     end
 end)
```
Cooldude2606, line 209: Can you explain why you need to do this? What is wrong with ordering the deconstruction in the logic above?

Cooldude2606 (changes requested):

modules/addons/miner.lua
```diff
@@ -1,64 +1,183 @@
...
-                end
+    local pipe_build = {}
 
-                if entity.has_flag('not-deconstructable') then
-                    -- if it can deconstruct
-                    return
-                end
+    if entity.fluidbox and #entity.fluidbox > 0 then
+        -- if require fluid to mine
+        if config.fluid then
```
Cooldude2606, line 107: Combine these into a single condition. With `config.fluid` first.

modules/addons/miner.lua
Cooldude2606: No resolved, no change was made and no explaintion given. This check should be centred on the target with a radius of 1. Is there a reason that this is not the case?

modules/addons/miner.lua
```diff
@@ -1,64 +1,183 @@
...
+
+    return false
+end
+
+local function chest_check(entity)
+    local target = drop_target(entity)
+
+    if chest_entity(entity) then
+        return
+    end
```
Cooldude2606, line 60: Should this not be `chest_entity(target)` because we need to check that the target can be removed.

modules/addons/miner.lua
Cooldude2606: NB: This is required because if a miner is marked for deconstruction in the same tick that it's resources are depleted it does not output the last ore. Instead it holds onto it in an internal inventory.

Cooldude2606 (approved):

PR #275, opened 2024-01-29
Lawnmower fix

PHIDIAS0303:


Cooldude2606 (approved):
The condition could be written better, but it works.

Issue #101, opened 2019-06-04
Adaptive Pollution Ageing


flbuddymooreiv:
**Feature overview**
Name: Adaptive Pollution Ageing
TL;DR: Scale `game.map_settings.pollution.ageing` from 1 to 10 based on game progress to alleviate the pathfinder and active entity counts for the server and all clients

**Details**
Bigger base = more pollution. More pollution = more pollution-biter interface. More angry biters = more active entities. At some point, you have obviously got enough resources and defenses to defend against any attack. So spawners do help provide a challenge during expansion, but the retaliation is nothing more than a UPS drain as you're waiting for unit groups to attack as the pathfinder can only handle so much routing. Thousands of groups could take a day or more to all attack and inevitably be vaporized.

**Possible methods**
Linearly interpolate ageing from 1 to 10 over rocket launches 1 to 1000. I.e. 0,1 rocket launches = pollution ageing 1, 1000 launches = pollution ageing 10.

**Additional context**
This doesn't make environmental or logical sense, but by the time you've launched 1000 rockets, the next step is to "clear the pollution cloud" which plainly put can't really happen in a multiplayer server on a week long reset, purely due to pollution. It's sad.

PR #277, opened 2024-01-29
Adaptive Pollution Ageing

PHIDIAS0303:
#101

### DIFF
--- a/modules/commands/research.lua
+++ b/modules/commands/research.lua
@@ -43,7 +43,7 @@ local function research_notification(event)
     end
 
     if is_inf_res then
-        if event.research.name == 'mining-productivity-4' and event.research.force.technologies['mining-productivity-4'].level > 4 then
+        if event.research.name == 'mining-productivity-4' and event.research.level > 4 then
             if config.bonus.enabled then
                 event.research.force[config.bonus.name] = base_rate + event.research.level * config.bonus.rate
             end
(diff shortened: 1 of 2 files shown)


modules/commands/research.lua
```diff
@@ -43,10 +43,14 @@ local function research_notification(event)
     end
 
     if is_inf_res then
-        if config.bonus.enabled then
-            if event.research.name == 'mining-productivity-4' and event.research.force.technologies['mining-productivity-4'].level > 4 then
+        if event.research.name == 'mining-productivity-4' and event.research.force.technologies['mining-productivity-4'].level > 4 then
```
Cooldude2606, line 46: You can index `level` directly from `event.research` because you have checked that the name is mining prod 4.

modules/commands/research.lua
```diff
@@ -43,10 +43,14 @@ local function research_notification(event)
...
+            if config.bonus.enabled then
                 event.research.force[config.bonus.name] = base_rate + event.research.force.technologies['mining-productivity-4'].level * config.bonus.rate
             end
+
+            if config.pollution_ageing_by_research then
+                game.map_settings.pollution.ageing = math.min(10, (event.research.force.technologies['mining-productivity-4'].level - 1) / 5)
```
Cooldude2606, line 52: Same as above; but also this can trigger when `level = 5` which results in `(5 - 1) / 5` which is `0.8` which is lower than the default ageing of `1` which means it decreases ageing on the first research. This should not happen.

Cooldude2606 (changes requested):

modules/commands/research.lua
Cooldude2606: Line 46 still needs to be updated.

Cooldude2606 (approved):

Issue #279, opened 2024-02-03
Spectate after Follow embodies you


DrahcPro:
**Factorio version**
Version number: 1.1.101

**Scenario version**
Version number: whatever s3 is on currently

**Bug description**
When I /follow somebody and want to use /spectate after it kills me and leaves the body I left when going into /follow forever

**Reproduction method**
1. /follow {a person}
2. /spectate
3. Enjoy your extra body

**Expected behavior**
To be put back into my own body

**Screenshots**
![image](https://github.com/explosivegaming/scenario/assets/37268586/97aabd37-55eb-4b6b-979d-9483082dbdf6)

**Additional context**
None, was testing donator perks

PR #280, opened 2024-02-03
spectate check

PHIDIAS0303:
#279


DrahcPro:
Like I have shown before I am unfamiliar with the data structures we use, but isn't there a flag for when a player is ghosting and or a general function for stopping a ghost feature?

Cooldude2606:
This PR does not fix the underlying issue, see #285

PR #283, opened 2024-02-04
Playerdata Possible NA fix

PHIDIAS0303:


Cooldude2606 (changes requested):
You removed `format_time_short` and its `*3600` so now the units passed to `format_time` are incorrect. This needs to be fixed.

PR #286, opened 2024-02-12
Vlayer

PHIDIAS0303:
it uses more than a few million landfills in this run. it is a bit too much considering you will build solar on land more than on water usually

### DIFF
--- a/modules/commands/vlayer.lua
+++ b/modules/commands/vlayer.lua
@@ -28,8 +28,6 @@ Commands.new_command('vlayer-info', 'Vlayer Info')
     local c = vlayer.get_circuits()
 
     for k, v in pairs(c) do
-        player.print(v .. ' : ' .. k)
+        Commands.print(v .. ' : ' .. k)
     end
-
-    return Commands.success
 end)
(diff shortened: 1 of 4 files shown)


Cooldude2606 (changes requested):
Could you also fix `starting_value` of accumulators not being 0.

modules/commands/vlayer.lua
```diff
@@ -15,10 +15,21 @@ Commands.new_command('personal-battery-recharge', 'Recharge Player Battery upto
...
 
     return Commands.success
 end)
+
+Commands.new_command('vlayer-info', 'Vlayer Info')
+:register(function(player)
+    local c = vlayer.get_circuits()
+
+    for k, v in pairs(c) do
+        player.print(v .. ' : ' .. k)
```
Cooldude2606, line 31: Use `Commands.print`

Cooldude2606 (changes requested):
There are better solutions than adding a new command. But this will work for now. Will merge once you change it to use `Commands.print`

Cooldude2606 (dismissed):

Cooldude2606 (approved):

PR #287, opened 2024-02-13
Personal Logistic

PHIDIAS0303:
0 should be trash all instead


Cooldude2606 (approved):
There is an additional change made here that wasn't described but I am fine with it.

PR #290, opened 2024-02-18
Personal Battery Recharge

PHIDIAS0303:


modules/commands/vlayer.lua
```diff
@@ -8,13 +8,19 @@ local vlayer = require 'modules.control.vlayer'
...
+    end
+
     local armor = player.get_inventory(defines.inventory.character_armor)[1].grid
 
     for i=1, #armor.equipment do
         local target = math.floor(armor.equipment[i].max_energy * amount)
 
         if armor.equipment[i].energy < target then
-            local energy_required = math.min(math.floor(target - armor.equipment[i].energy), vlayer.get_statistics()['energy_storage'])
+            local energy_required = math.min(math.floor(target - armor.equipment[i].energy), stat['energy_storage'])
```
Cooldude2606, line 23: This will not work because `energy_storage` is not reduced. This may result in it becoming negative when charging multiple items.

Cooldude2606 (changes requested):

modules/commands/vlayer.lua
```diff
@@ -8,6 +8,10 @@ local vlayer = require 'modules.control.vlayer'
 Commands.new_command('personal-battery-recharge', 'Recharge Player Battery upto a portion with vlayer')
 :add_param('amount', 'number-range', 0.2, 1)
 :register(function(player, amount)
+    if vlayer.get_statistics()['energy_sustained'] == 0 then
+        return Commands.print('vlayer need to be running to get this command work')
```
Cooldude2606, line 12: Must be `Commands.error`

PR #291, opened 2024-02-22
vlayer auto connect circuit

PHIDIAS0303:


Cooldude2606 (changes requested):
1) This code should not be in the GUI module, put it in the control module.
2) `next(res.circuit[k]) ~= nil` is not needed.
3) You have repeated the same code three times, make it a function.
Other than that, the feature works as intended and there are no bugs +1

modules/control/vlayer.lua
```diff
@@ -269,14 +269,24 @@ end
...
     local interface = surface.create_entity{name='logistic-chest-storage', position=position, force='neutral'}
     table.insert(vlayer_data.entity_interfaces.storage_input, interface)
 
     if last_user then
         interface.last_user = last_user
     end
 
+    if circuit then
+        for k, _ in pairs(circuit) do
+            if next(circuit[k]) ~= nil then
```
Cooldude2606, line 282: `if next(circuit[k]) ~= nil then` is not needed.

PR #296, opened 2024-06-05
vlayer

PHIDIAS0303:


modules/gui/vlayer.lua
```diff
@@ -59,7 +72,7 @@ local function vlayer_convert_chest(player)
     local circuit = entity.circuit_connected_entities
 
     if (not entity.get_inventory(defines.inventory.chest).is_empty()) then
-        player.print('Chest is not emptied')
+        player.print{'vlayer.steel-chest-detect'}
```
Cooldude2606, line 75:
```suggestion
        player.print{'vlayer.steel-chest-empty'}
```

modules/gui/vlayer.lua
```diff
@@ -199,6 +216,29 @@ Gui.element{
...
 }
 
+--- Display label for the remaining surface area
+-- @element vlayer_gui_display_signal_remaining_surface_area_name
+local vlayer_gui_display_signal_remaining_surface_area_name =
+Gui.element{
+    type = 'label',
+    name = 'vlayer_display_signal_remaining_surface_area_name',
+    caption = {'vlayer.display-remaining_surface_area'},
+    tooltip = {'vlayer.display-remaining_surface_area-tooltip'},
```
Cooldude2606, line 226: Should use `-` not `_`

modules/gui/vlayer.lua
```diff
@@ -222,116 +264,151 @@ Gui.element(function(_, parent, name)
...
-    local res = vlayer_convert_chest(player)
+}:on_click(function(_, element, _)
+    local target = element.parent[vlayer_gui_control_type.name].selected_index
+    local full_list = {}
 
-    if res then
-        vlayer.create_output_interface(player.surface, res.pos, res.circuit, player)
-        game.print(player.name .. ' built a vlayer storage output on ' .. pos_to_gps_string(res.pos))
+    for i=1, vlayer.get_interface_counts()[vlayer_control_type_list[target]], 1 do
+        table.insert(full_list, i)
```
Cooldude2606, line 305: Would be better if this was coordinates rather than just a number.

modules/gui/vlayer.lua
```diff
@@ -222,116 +264,151 @@ Gui.element(function(_, parent, name)
...
-
-    local interfaces = vlayer.get_interface_counts()
-    game.print(player.name .. ' removed a vlayer ' .. interface_type .. ' on ' .. pos_to_gps_string(interface_position))
-    element.parent[vlayer_gui_control_storage_input.name].enabled = (interfaces.storage_input < config.interface_limit.storage_input)
-    element.parent[vlayer_gui_control_storage_output.name].enabled = (interfaces.storage_output < config.interface_limit.storage_output)
-    element.parent[vlayer_gui_control_circuit.name].enabled = (interfaces.circuit < config.interface_limit.circuit)
-    element.parent[vlayer_gui_control_power.name].enabled = (interfaces.energy < config.interface_limit.energy)
+    local target = element.parent[vlayer_gui_control_type.name].selected_index
+    local n = element.parent[vlayer_gui_control_list.name].selected_index
+    local interface_type, interface_position = vlayer.remove_interface(vlayer.get_interfaces()[vlayer_control_type_list[target]][n].surface, vlayer.get_interfaces()[vlayer_control_type_list[target]][n].position)
```
Cooldude2606, line 410: Put `vlayer.get_interfaces()[vlayer_control_type_list[target]` into a variable.

modules/control/vlayer.lua
```diff
@@ -580,21 +606,40 @@ local function handle_energy_interfaces()
...
+        for k, v in pairs(vlayer_data.storage.power_items) do
+            if v.count > 0 then
+                local to_burn
+
+                if (v.count * v.value) < max_burning then
+                    to_burn = v.count
+
+                else
+                    to_burn = math.min(max_burning / v.value)
+                end
```
Cooldude2606, line 623:
```suggestion
                to_burn = math.min(v.count, max_burning / v.value)
```

modules/control/vlayer.lua
```diff
@@ -580,21 +606,40 @@ local function handle_energy_interfaces()
...
+
+                vlayer_data.storage.energy = vlayer_data.storage.energy + (to_burn * v.value)
```
Cooldude2606, line 625: You also need to update `max_burning` so that the next item does not produce too much

locale/en/gui.cfg
```diff
@@ -210,7 +210,34 @@ disabled=disabled
...
+display-sustained-production=[virtual-signal=signal-S] Sustained Production
+display-max-capacity=[virtual-signal=signal-C] Battery Capacity
+display-current-capacity=[virtual-signal=signal-E] Battery Charge
+display-remaining_surface_area=[virtual-signal=signal-R] Remaining Surface
+display-item-solar-tooltip=The amount of Solar Panel
+display-item-accumulator-tooltip=The amount of Accumulator
+display-current-production-tooltip=Current Power Production
+display-sustained-production-tooltip=Sustained Power Production
+display-max-capacity-tooltip=Battery Max Capacity
+display-current-capacity-tooltip=Battery Current Charge
```
Cooldude2606, line 226: Would help a lot to have units listed here, eg MJ or MW.

locale/en/gui.cfg
```diff
@@ -210,7 +210,34 @@ disabled=disabled
 toggle-msg=Fast decon has been __1__
 
 [vlayer]
-main-tooltip=Enable Vlayer GUI
+main-tooltip=vlayer GUI
```
Cooldude2606, line 213:
```suggestion
main-tooltip=Vlayer
```

config/vlayer.lua
```diff
@@ -55,15 +55,22 @@ return {
...
+            fuel_value = 2, -- MJ
+            power = true -- turn all wood to power to reduce trash
         },
+        ['coal'] = {
+            starting_value = 0,
+            required_area = 0,
+            surface_area = 0,
+            fuel_value = 4, -- MJ
+            power = true -- turn all coal to power to reduce trash
+        }
```
Cooldude2606, line 72: Coal should not be here by default, it is not trash and must be intentionally mined. The amount gained from large rocks is not sufficient when it can be easily abused.

Cooldude2606 (changes requested):

modules/gui/vlayer.lua
Cooldude2606: Line 114 has not been changed.

locale/en/gui.cfg
Cooldude2606: Should be "V" not "v" to match style of other GUIs.

modules/control/vlayer.lua
```diff
@@ -580,21 +606,34 @@ local function handle_energy_interfaces()
     -- Cap the stored energy to the allowed capacity
     if not config.unlimited_capacity and vlayer_data.storage.energy > vlayer_data.properties.capacity * mega then
         vlayer_data.storage.energy = vlayer_data.properties.capacity * mega
+
+    -- burn the trash to produce power
+    elseif vlayer_data.storage.power_items then
+        local max_burning
```
Cooldude2606, line 612: This does not need to be in this scope.

modules/gui/vlayer.lua
```diff
@@ -222,116 +264,153 @@ Gui.element(function(_, parent, name)
...
-    local interfaces = vlayer.get_interface_counts()
-    game.print(player.name .. ' removed a vlayer ' .. interface_type .. ' on ' .. pos_to_gps_string(interface_position))
-    element.parent[vlayer_gui_control_storage_input.name].enabled = (interfaces.storage_input < config.interface_limit.storage_input)
-    element.parent[vlayer_gui_control_storage_output.name].enabled = (interfaces.storage_output < config.interface_limit.storage_output)
-    element.parent[vlayer_gui_control_circuit.name].enabled = (interfaces.circuit < config.interface_limit.circuit)
-    element.parent[vlayer_gui_control_power.name].enabled = (interfaces.energy < config.interface_limit.energy)
+    local target = element.parent[vlayer_gui_control_type.name].selected_index
+    local n = element.parent[vlayer_gui_control_list.name].selected_index
+    local t = vlayer.get_interfaces()[vlayer_control_type_list[target]]
+    local interface_type, interface_position = vlayer.remove_interface(t[n].surface, t[n].position)
```
Cooldude2606, line 412: `attempt to index field '?' (a nil value)`
Need a nil check here, or a length check on t against n.

Cooldude2606 (changes requested):
Removing an interface does not refresh the dropdown which leads to unexpected behaviour. I would recommend creating a custom event for this if you understand how. If not making refresh into a function will be sufficient and having it called for all players when adding or removing an interface.

locale/en/gui.cfg
PHIDIAS0303: that dont look good

modules/gui/vlayer.lua
```diff
@@ -222,116 +264,169 @@ Gui.element(function(_, parent, name)
...
+    local target = table[vlayer_gui_control_type.name].selected_index
+    local full_list = {}
 
-    element.enabled = (vlayer.get_interface_counts().storage_input < config.interface_limit.storage_input)
-end)
+    if target then
+        local interface = vlayer.get_interfaces()[vlayer_control_type_list[target]]
+
+        for i=1, vlayer.get_interface_counts()[vlayer_control_type_list[target]], 1 do
+            table.insert(full_list, i .. ' X ' .. interface[i].position.x .. ' Y '.. interface[i].position.y)
```
Cooldude2606, line 301: `attempt to call field 'insert' (a nil value)`
Do not use `table` as a variable name on line 293.

modules/gui/vlayer.lua
```diff
@@ -364,11 +459,13 @@ Gui.element(function(definition, parent)
...
 
-    table[vlayer_gui_control_storage_input.name].visible = visible
-    table[vlayer_gui_control_storage_output.name].visible = visible
-    table[vlayer_gui_control_circuit.name].visible = visible
-    table[vlayer_gui_control_power.name].visible = visible
+    table[vlayer_gui_control_type.name].visible = visible
+    table[vlayer_gui_control_list.name].visible = visible
+    table[vlayer_gui_control_refresh.name].visible = visible
+    table[vlayer_gui_control_see.name].visible = visible
+    table[vlayer_gui_control_build].visible = visible
```
Cooldude2606, line 466:
```suggestion
    table[vlayer_gui_control_build.name].visible = visible
```

Cooldude2606 (changes requested):
Switching interface type should also call `vlayer_gui_List_refresh`

modules/control/vlayer.lua
```diff
@@ -580,21 +606,32 @@ local function handle_energy_interfaces()
     -- Cap the stored energy to the allowed capacity
     if not config.unlimited_capacity and vlayer_data.storage.energy > vlayer_data.properties.capacity * mega then
         vlayer_data.storage.energy = vlayer_data.properties.capacity * mega
+
+    -- burn the trash to produce power
+    elseif vlayer_data.storage.power_items then
+        for k, v in pairs(vlayer_data.storage.power_items) do
+            local max_burning = (vlayer_data.properties.capacity * mega / 2) - vlayer_data.storage.energy
+
+            if v.count > 0 then
```
Cooldude2606, line 615:
```suggestion
            if v.count > 0 and max_burning > 0 then
```
This change is needed otherwise when the stored energy is above 50% it will start consuming power to create wood.

Cooldude2606 (approved):

PR #300, opened 2024-07-09
Bonus GUI

PHIDIAS0303:


modules/data/bonus.lua
```diff
@@ -30,12 +30,12 @@ local function apply_bonus(player, stage)
...
-            if stage == 0 then
-                player[v.name] = v.min
-            else
-                player[v.name] = v.min + (v.max - v.min) * stage / 10
+    for k, v in pairs(config.player_bonus) do
+        player[k] = v.value * stage / 10
+
+        if v.combined_bonus then
+            for i=1, #v.combined_bonus, 1 do
+                player[v.combined_bonus[i]] = v.value * stage / 10
```
Cooldude2606, line 38: From my understanding that `value` is the start / base value, this math would not work. Are you able to clarify this.

locale/en/gui.cfg
```diff
@@ -209,6 +209,47 @@ enabled=enabled
...
 
+[bonus]
+main-tooltip=Bonus
+control-pts-a=Points available
+control-pts-n=Points needed
+control-pts-r=Points remaining
+control-refresh=Refresh
+control-apply=Apply
+control-pts-exceed=Points allocated exceeded allowance
+display-cmms=CMMS
```
Cooldude2606, line 220: Rather than using acronyms, shorthands will be better such as `Mining` `Running` `Inventory` etc

Cooldude2606 (changes requested):
Some changes needed; although I'll also comment on discord as some things can be done to make this much easier on yourself.
Also don't use `tonumber(element.caption)` when you can instead get the value directly from the slider.

locale/en/gui.cfg
PHIDIAS0303: it is too long

modules/data/bonus.lua
PHIDIAS0303: the value is just the bonus 10 in the old days.

config/bonus.lua
```diff
@@ -3,221 +3,271 @@
...
-            name = 'character_health_bonus',
-            enabled = true,
-            min = 0,
-            max = 200
+        ['character_mining_speed_modifier'] = {
+            value = 2,
+            max = 4,
+            scale = 0.25,
+            cost_scale = 1,
+            cost = 10
```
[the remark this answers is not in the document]
PHIDIAS0303: value is the default bonus 10 of the old days. max allow it to be changed even more. or just less if player don't need it.
the scale is the minimum of how much you can move.
the cost scale is how much points needed, each cost can give you how many cost scale

locale/en/gui.cfg
Cooldude2606: Could you explain why you are opposed to using more readable names?

PHIDIAS0303: they are just not updated i think

modules/gui/bonus.lua
```diff
@@ -0,0 +1,286 @@
...
+    type = 'button',
+    name = Gui.unique_static_name,
+    caption = {'bonus.control-refresh'}
+}:style{
+    width = config.gui_display_width['half']
+}:on_click(function(player, element, _)
+    local r = bonus_gui_pts_needed(player)
+    element.parent[bonus_gui_control_pts_n_count.name].caption = r
+    element.parent[bonus_gui_control_pts_r_count.name].caption = tonumber(element.parent[bonus_gui_control_pts_a_count.name].caption) - r
+end)
```
Cooldude2606, line 140: Because the values update automatically, it would be a lot more useful for this button to reset the sliders to the players current bonus. (Kind of like a 'cancel changes' button)

modules/gui/bonus.lua
```diff
@@ -0,0 +1,286 @@
...
+    local disp = Gui.scroll_table(bonus_set, 360, 2, 'disp')
+
+    bonus_gui_control_pts_a(disp)
+    bonus_gui_control_pts_a_count(disp)
+
+    bonus_gui_control_pts_n(disp)
+    bonus_gui_control_pts_n_count(disp)
+
+    bonus_gui_control_pts_r(disp)
+    bonus_gui_control_pts_r_count(disp)
```
Cooldude2606, line 176: Should set the two count labels here rather than needing to manually refresh when you join.

modules/gui/bonus.lua
```diff
@@ -0,0 +1,286 @@
...
+        allow_decimal = false,
+        allow_negative = false,
+        style = 'heading_1_label'
+    }
+    count.style.width = config.gui_display_width['count']
+
+    return slider
+end)
+:on_value_changed(function(player, element, _event)
+    element.parent[element.tags.counter].caption = element.slider_value
```
Cooldude2606, line 225: Would be nice if the percentage based bonuses (Mining/Running/Crafting) showed as `150%` rather than `1.5`. This can be done by adding a `is_percentage` flag from the config into `element.tags` on line 205. Then you can test for it here.

modules/gui/bonus.lua
```diff
@@ -0,0 +1,286 @@
...
+    slider.style.width = config.gui_display_width['slider']
+    slider.style.horizontally_stretchable = true
+
+    local count = parent.add{
+        type = 'label',
+        name = name .. '_count',
+        caption = bonus.value,
+        numeric = true,
+        allow_decimal = false,
+        allow_negative = false,
```
Cooldude2606, line 217: These do nothing for a label, they are used to control allowed user input on text fields. They can be removed.

Cooldude2606 (changes requested):
See discord for suggested changes to the config file.

locale/en/gui.cfg
```diff
@@ -162,18 +162,18 @@ servers-tab=Servers
...
-servers-6=S6 Donator
-servers-d6=This is our donator only server, online when requested.
+servers-3=S3 Off
+servers-d3=This server is closed.
+servers-4=S4 Off
+servers-d4=This server is closed.
+servers-5=S5 Off
+servers-d5=This server is closed.
+servers-6=S6 Off
+servers-d6=This server is closed.
```
Cooldude2606, line 176: Why this change?

locale/en/commands.cfg
```diff
@@ -53,11 +53,6 @@ unavailable=They was a problem getting you to spawn, please try again later.
 [expcom-repair]
 result=__1__ entites were revived and __2__ were healed to max health.
 
-[expcom-bonus]
-set=Your bonus has been set to __1__.
-perm=You dont have enough permission to set more than __1__.
-wip=This command is temporary and will be replaced at some point in the future.
-
```
Cooldude2606, line 60: While the command file still exists, these also need to remain.

Cooldude2606 (changes requested):
`on_player_died` needs to be added for instant respawn.
`on_player_respawned` to reapply bonus, make sure to check remaining >= 0.

locale/en/gui.cfg
PHIDIAS0303: because i pasted my server one

Cooldude2606 (approved):
All code looks good to me. However, since this impacts the rewards for donators I will need to inform them before this is merged.

PR #301, opened 2024-07-09
Module GUI

PHIDIAS0303:


modules/gui/module.lua
```diff
@@ -105,66 +97,77 @@ end
 Selection.on_selection(SelectionModuleArea, function(event)
     local area = aabb_align_expand(event.area)
     local player = game.get_player(event.player_index)
-
-    if player == nil then
-        return
-    end
-
     local frame = Gui.get_left_element(player, module_container)
+    local table = frame.container.scroll.table
```
Cooldude2606, line 101: Don't use local name `table` because this shadows the table library. I normally use `tbl` when this happens, although anything more descriptive also works.

modules/gui/module.lua
```diff
@@ -69,33 +74,20 @@ end
 
 local function apply_module(player, area, machine, modules)
     for _, entity in pairs(player.surface.find_entities_filtered{area=area, name=machine, force=player.force}) do
-        if config.machine_craft[machine] then
```
Cooldude2606, line 72: This condidtion can not be removed because not all machines that accept modules have recipes. eg mining drill.

Cooldude2606 (changes requested):

modules/gui/module.lua
```diff
@@ -69,33 +74,24 @@ end
 
 local function apply_module(player, area, machine, modules)
     for _, entity in pairs(player.surface.find_entities_filtered{area=area, name=machine, force=player.force}) do
+        local m_current_recipe
+
         if config.machine_craft[machine] then
```
Cooldude2606, line 79: Since you removed `machine_craft` from the config, you need to use `machine.prototype.crafting_speed` (which is nil when the entity is not a crafting machine)

modules/gui/module.lua
```diff
@@ -105,66 +101,77 @@ end
...
         end
+
     else
         local mf = elem_filter.normal
 
-        if config.machine_prod_disallow[element.elem_value] ~= nil then
-            if config.machine_prod_disallow[element.elem_value] then
-                mf = elem_filter.prod
-            end
+        if not config.machine[element.elem_value].prod then
```
Cooldude2606, line 167: `element.elem_value` is undefined.

modules/gui/module.lua
```diff
@@ -105,66 +101,77 @@ end
...
+            mf = elem_filter.prod
         end
 
         for i=1, config.module_slot_max do
-            frame.container.scroll.table[element .. i].enabled = true
-            frame.container.scroll.table[element .. i].elem_filters = mf
-            frame.container.scroll.table[element .. i].elem_value = nil
+            scroll_table[element .. i].enabled = true
```
Cooldude2606, line 172: Should this not be `false` since there is no selected entity?

modules/gui/module.lua
```diff
@@ -105,66 +101,77 @@ end
...
+
             else
-                frame.container.scroll.table[element .. i].enabled = false
-                frame.container.scroll.table[element .. i].elem_value = nil
+                scroll_table[element .. i].enabled = false
+                scroll_table[element .. i].elem_value = nil
             end
-            frame.container.scroll.table[element .. i].elem_filters = elem_filter.normal
+
+            scroll_table[element .. i].elem_filters = elem_filter.normal
```
Cooldude2606, line 161: We need a check for `config.machine[scroll_table[element .. '0'].elem_value].prod` so you can't put prod modules into beacons.

Cooldude2606 (changes requested):
Sorry found some more bugs in my final test of the module.

modules/gui/module.lua
```diff
@@ -105,66 +112,78 @@ end
...
+    local scroll_table = frame.container.scroll.table
 
-    if frame.container.scroll.table[element .. '0'].elem_value ~= nil then
+    if scroll_table[element .. '0'].elem_value then
         for i=1, config.module_slot_max do
-            if i <= game.entity_prototypes[frame.container.scroll.table[element .. '0'].elem_value].module_inventory_size then
-                frame.container.scroll.table[element .. i].enabled = true
-                frame.container.scroll.table[element .. i].elem_value = config.machine[frame.container.scroll.table[element .. '0'].elem_value]
+            if i <= game.entity_prototypes[scroll_table[element .. '0'].elem_value].module_inventory_size then
+                if config.machine[scroll_table[element .. '0'].elem_value].prod then
```
Cooldude2606, line 164: This condition (or the filters) need to be inverted. Currently `prod` is true when prod modules are supported but the `prod` filter excludes prod modules.

Cooldude2606 (changes requested):
Filters still are not correct, you need to flip "normal" and "prod" not remove the invert flag, that part was correct.

Cooldude2606 (approved):

PR #303, opened 2024-07-11
VLAYER

PHIDIAS0303:


Cooldude2606:
This change can not be made because `surface.daytime` can be modified from other scripts.
For example if `surface.always_day` or `surface.freeze_daytime` are true then `time` will continue to tick when it should be constant.

PHIDIAS0303:
but the main thing is that I need the tick to do the day-night calculation of solar power ...

Cooldude2606:
> but the main thing is that I need the tick to do the day-night calculation of solar power ...

For it to be accurate for vlayer solar it must use `surface.daytime` because this is used to calculate solar power production not game tick.

modules/control/vlayer.lua
```diff
@@ -440,12 +440,14 @@ function vlayer.get_statistics()
         total_surface_area = vlayer_data.properties.total_surface_area,
         used_surface_area = vlayer_data.properties.used_surface_area,
         remaining_surface_area = math.max(vlayer_data.properties.total_surface_area - vlayer_data.properties.used_surface_area, 0),
+        multiplier = get_production_multiplier(),
```
Cooldude2606, line 443: `multiplier` is vauge, use full name `production_multiplier`

modules/control/vlayer.lua
```diff
@@ -440,12 +440,14 @@ function vlayer.get_statistics()
...
         energy_production = vlayer_data.properties.production * mega * get_production_multiplier(),
         energy_sustained = vlayer_data.properties.production * mega * get_sustained_multiplier(),
         energy_capacity = vlayer_data.properties.capacity * mega,
         energy_storage = vlayer_data.storage.energy,
         day = math.floor(game.tick / vlayer_data.surface.ticks_per_day),
-        time = math.floor(vlayer_data.surface.daytime * vlayer_data.surface.ticks_per_day)
+        day_length = vlayer_data.surface.ticks_per_day,
+        time = math.floor(game.tick % vlayer_data.surface.ticks_per_day)
```
Cooldude2606, line 450: Unable to accept the use of `game.tick` when it introduces inconsistencies. Add a `tick` property and combine with your `day_length` property in combinators if you absolutely must.

Cooldude2606 (changes requested):

Cooldude2606 (changes requested):
Not what I meant why I said add a tick property. I meant a `tick` signal that is equal to `game.tick` so that you can do `tick % day_length` in a combinator and return `time` to be based on `daytime`.

modules/gui/vlayer.lua
```diff
@@ -10,8 +10,20 @@ local Event = require 'utils.event' --- @dep utils.event
 local format_number = require('util').format_number --- @dep util
 local config = require 'config.vlayer' --- @dep config.vlayer
 local vlayer = require 'modules.control.vlayer'
+local Selection = require 'modules.control.selection' --- @dep modules.control.selection
+local SelectionConvertArea = 'ConvertArea'
```
Cooldude2606, line 14: This must be a unique string. Currently it conflicts with waterfill.
```suggestion
local SelectionConvertArea = 'VlayerConvertChest'
```

modules/gui/vlayer.lua
```diff
@@ -59,26 +71,67 @@ local function format_energy(amount, unit)
...
+    local disp = frame.container['vlayer_st_2'].disp.table
+    local target = vlayer_control_type_list[disp[vlayer_gui_control_type.name].selected_index]
+
+    if #entities == 0 then
         player.print{'vlayer.steel-chest-detect'}
-        return
+        return nil
+
+    elseif #entities > 1 then
+        player.print{'vlayer.result-unable', {'vlayer.control-type-' .. target:gsub('_', '-')}}
```
Cooldude2606, line 93: Can you add a reason to this, eg `Unable to build Energy interface: Multiple steel chests selected`
Also add a reason to the other places this is used. ie `At interface limit` and `Not enough room`

Cooldude2606 (changes requested):
I really like the change to use the selection tool. It all works correctly but the error messages need improving.

Cooldude2606 (approved):

PR #304, opened 2024-07-11
Locale

PHIDIAS0303:


Cooldude2606 (approved):
Didn't find any syntax errors. And I just have to trust that these translations are accurate.

PR #310, opened 2024-08-02
Copy Paste Machine Setting

PHIDIAS0303:


modules/gui/module.lua
```diff
@@ -250,3 +250,58 @@ Event.add(defines.events.on_gui_elem_changed, function(event)
...
+    local destination = event.destination
+    local player = game.players[event.player_index]
+
+    if not player then
+        return
+    end
+
+    if not source then
+        return
+    end
```
Cooldude2606, line 265: Needs to be `if not source or not source.valid then`

modules/gui/module.lua
```diff
@@ -250,3 +250,58 @@ Event.add(defines.events.on_gui_elem_changed, function(event)
...
+
+    local source_inventory_content = source_inventory.get_contents()
+
+    if not source_inventory_content then
+        return
+    end
+
+    if not destination then
+        return
+    end
```
Cooldude2606, line 281: Needs to be `if not destination or not destination.valid then`
Also this check should be done before accessing the inventory of source because it is a cheaper check.

modules/gui/module.lua
```diff
@@ -250,3 +250,58 @@ Event.add(defines.events.on_gui_elem_changed, function(event)
...
+
+    if not source_inventory then
+        return
+    end
+
+    local source_inventory_content = source_inventory.get_contents()
+
+    if not source_inventory_content then
+        return
+    end
```
Cooldude2606, line 277: These checks need to be after line 301 if you intend for direction copying to work.

modules/gui/module.lua
```diff
@@ -250,3 +250,58 @@ Event.add(defines.events.on_gui_elem_changed, function(event)
...
+
+    if not destination then
+        return
+    end
+
+    if (source.name == destination.name or source.prototype.fast_replaceable_group == destination.prototype.fast_replaceable_group) then
```
Cooldude2606, line 283: This needs a comment to describe that it copies direction on paste and is unrelated to the module code.

Cooldude2606 (changes requested):

Cooldude2606 (approved):

Cooldude2606 (approved):

PR #311, opened 2024-08-04
Lawnmower

PHIDIAS0303:
clear decoration when place down


modules/addons/lawnmower.lua
```diff
@@ -31,3 +32,25 @@ Commands.new_command('lawnmower', 'Clean up biter corpse, decoratives and nuclea
...
+    destroy_decoratives(event.created_entity, event.created_entity.surface_index, event.created_entity.selection_box)
+end)
+
+Event.add(defines.events.script_raised_built, function(event)
+    destroy_decoratives(event.entity, event.entity.surface_index, event.entity.selection_box)
+end)
+
+Event.add(defines.events.script_raised_revive, function(event)
+    destroy_decoratives(event.entity, event.entity.surface_index, event.entity.selection_box)
+end)
```
Cooldude2606, line 56: Why `surface_index` and not `surface` ?

Cooldude2606 (commented):
Going to ask if this is something people want; as personally I like the decorations.

Cooldude2606:
The vote was for the default to be false, either change it or provide evidence of significant reduction in file size.

Cooldude2606 (approved):

PR #312, opened 2024-08-06
Production GUI

PHIDIAS0303:
a quick lookup to see how to improve current production. also not blocking the view


Cooldude2606:
Luacheck has been updated on dev, please merge changes to fix ci checks.

modules/gui/production.lua
```diff
@@ -0,0 +1,206 @@
...
+-- @element production_data_group
+local production_data_group =
+Gui.element(function(_definition, parent, i)
+    local item = parent.add{
+        type = 'choose-elem-button',
+        name = 'production_' .. i .. '_e',
+        elem_type = 'item',
+        style = 'slot_button'
+    }
+    item.style.width = 80
```
Cooldude2606, line 105: Slots should be square, setting the width like this stretches it.

modules/gui/production.lua
```diff
@@ -0,0 +1,206 @@
...
+    selected_index = 3
+}:style{
+    width = 80
+}
+
+local data_1s =
+Gui.element{
+    type = 'label',
+    name = 'production_0_1',
+    caption = '+',
```
Cooldude2606, line 48: Can you change this to be "Made", "Used", "Sum". There is too much empty space at the moment.

modules/gui/production.lua
```diff
@@ -0,0 +1,206 @@
...
+            local item = table[production_prefix .. '_e'].elem_value
+
+            if item then
+                local add = math.floor(stat.get_flow_count{name=item, input=true, precision_index=precision_value, count=false} / 6) / 10
+                local minus = math.floor(stat.get_flow_count{name=item, input=false, precision_index=precision_value, count=false} / 6) / 10
+                local sum = add - minus
+
+                table[production_prefix .. '_1'].caption = format_n(add)
+                table[production_prefix .. '_2'].caption = format_n(minus)
+                table[production_prefix .. '_3'].caption = format_n(sum)
```
Cooldude2606, line 189: Units are not mentioned anywhere, it should at least be in a tooltip or the blue "info" icon that some of the guis have.

modules/gui/production.lua
```diff
@@ -0,0 +1,206 @@
...
+}
+
+--- A vertical flow containing all the production control
+-- @element production_control_set
+local production_control_set =
+Gui.element(function(_, parent, name)
+    local production_set = parent.add{type='flow', direction='vertical', name=name}
+    local disp = Gui.scroll_table(production_set, 368, 4, 'disp')
+
+    production_time_scale(disp)
```
Cooldude2606, line 87: If you dislike the difference between the sqaure slots and this dropdown, then you could have an attempt at putting it into the header similar to the `+` button on task list or warps.

modules/gui/production.lua
```diff
@@ -0,0 +1,206 @@
...
+    [1] = defines.flow_precision_index.five_seconds,
+    [2] = defines.flow_precision_index.one_minute,
+    [3] = defines.flow_precision_index.ten_minutes,
+    [4] = defines.flow_precision_index.one_hour,
+    [5] = defines.flow_precision_index.ten_hours
+}
+
+local font_color = {
+    [1] = {r = 0.3, g = 1, b = 0.3},
+    [2] = {r = 1, g = 0.3, b = 0.3}
```
Cooldude2606, line 20: Could you give these names such as "positive" and "negative".

Cooldude2606 (changes requested):
The GUI itself works great, no issues. Just a few style changes needed.

modules/gui/production.lua
```diff
@@ -0,0 +1,209 @@
...
+    return production_set
+end)
+
+production_container =
+Gui.element(function(definition, parent)
+    local container = Gui.container(parent, definition.name, 368)
+    Gui.header(container, {'production.main-tooltip'}, '', true)
+
+    production_control_set(container, 'production_st_1')
+    production_data_set(container, 'production_st_2')
```
Cooldude2606, line 163: These should be combined into one table because the titles do not align with the values.

Cooldude2606 (approved):

PR #315, opened 2024-08-15
Research

PHIDIAS0303:


modules/gui/research.lua
```diff
@@ -33,30 +33,37 @@ local empty_time = format_time(0, {
...
+}
+
 local res = {
 	['lookup_name'] = {},
 	['disp'] = {}
 }
 
-local res_total = 0
-local mi = 1
+local function research_init()
```
Cooldude2606, line 46: Looks like you have made this into a function and changed when it is calculated. This isn't needed and to prevent scope issues you can so `do end` blocks.
```suggestion
do
```

modules/gui/research.lua
```diff
@@ -93,49 +100,85 @@ local function research_res_n(res_)
...
 		end
     end
 end
 
+local function research_gui_update()
+	local res_disp = {}
+	local res_n = research_res_n(res['disp'])
+
+	for i=1, 8, 1 do
+		res_disp[i] = {}
```
Cooldude2606, line 144: You can move this into a variable to avoid all the indexing below. You can also set the defaults as the values in the else block because it is faster to add keys at creation than afterwards. eg
```lua
local next_res_disp = {
    name = '',
    target = '',
    -- etc
}
res_disp[i] = next_res_disp
-----
next_res_disp['name'] = res['disp'][res_i].name
next_res_disp['target'] = res['disp']
```

Cooldude2606 (changes requested):
Very happy with this PR. Only two issues are quite small ones.

modules/gui/research.lua
```diff
@@ -93,148 +101,184 @@ local function research_res_n(res_)
...
+	local res_disp = research_gui_update()
+
+	research_data_group(disp, 0)
+	disp['research_0_name'].caption = {'expcom-res.name'}
+	disp['research_0_target'].caption = {'expcom-res.target'}
+	disp['research_0_target'].style.horizontal_align = 'left'
+	disp['research_0_attempt'].caption = {'expcom-res.attempt'}
+	disp['research_0_attempt'].style.horizontal_align = 'left'
+	disp['research_0_difference'].caption = {'expcom-res.difference'}
+	disp['research_0_difference'].style.horizontal_align = 'left'
```
Cooldude2606, line 262: I think these look better when right aligned so that they line up with the data; what do you think?

Cooldude2606 (approved):

PR #318, opened 2024-08-27
Miner

PHIDIAS0303:
should be a slightly better method


modules/addons/miner.lua
```diff
@@ -156,14 +143,14 @@ Event.add(defines.events.on_resource_depleted, function(event)
         return
     end
 
-    local entities = event.entity.surface.find_entities_filtered{area={{event.entity.position.x - 1, event.entity.position.y - 1}, {event.entity.position.x + 1, event.entity.position.y + 1}}, type='mining-drill'}
+    local entities = event.entity.surface.find_entities_filtered{area={{event.entity.position.x - 3, event.entity.position.y - 3}, {event.entity.position.x + 3, event.entity.position.y + 3}}, type='mining-drill'}
```
Cooldude2606, line 146: Why was this changed to 3?

PHIDIAS0303: mining_drill_radius?

PHIDIAS0303:
will just change to this for now, the status is having some issue sometimes

modules/addons/miner.lua
Cooldude2606: We are searching a radius around an ore entity for a mining drill which can reach it. The value 1 is used because it is the difference between the mining radius and physical size of of the mining drill. This should remain as 1.

modules/addons/miner.lua
```diff
@@ -163,21 +152,6 @@ Event.add(defines.events.on_resource_depleted, function(event)
     end
 
     for _, entity in pairs(entities) do
-        if ((math.abs(entity.position.x - event.entity.position.x) < entity.prototype.mining_drill_radius) and (math.abs(entity.position.y - event.entity.position.y) < entity.prototype.mining_drill_radius)) then
```
Cooldude2606, line 166: This check is used to avoid the first call to `find_entities_filtered` within `miner_check`. Please add this condition back unless you can show it is faster to make the api call.

Cooldude2606 (commented):
This is required because if a miner is marked for deconstruction in the same tick that it's resources are depleted it does not output the last ore. Instead it holds onto it in an internal inventory.

_Originally posted by @Cooldude2606 in https://github.com/explosivegaming/scenario/pull/270#discussion_r1461006714_

We can not remove the `miner_data` table unless you have seen that the above issue has been fixed. If it has not been fixed then you haven't made any material changes in this PR.

modules/addons/miner.lua
```diff
@@ -64,7 +64,7 @@ local function chest_check(entity)
         return
     end
 
-    local radius = 2
```
Cooldude2606, line 67: We are searching around a chest, it makes no sence to use the mining drill radius here. Revert this change.

modules/addons/miner.lua
```diff
@@ -76,28 +76,26 @@ local function chest_check(entity)
     end
 
     if check_entity(target) then
-        table.insert(miner_data.queue, {t=game.tick + 10, e=target})
```
Cooldude2606, line 79: This should remove the chest (`target`) not the miner (`entity`) also the +10 is needed so that it is guaranteed the chest is deconstructed after the miner (which is +5). So revert all changes on this line.

Cooldude2606 (changes requested):

Cooldude2606 (approved):
Not sure why the `mining_target` check was removed, but I don't take issue with it since it makes the code simpler and its not clear if its beneficial.

PR #321, opened 2024-08-30
Surveillance

PHIDIAS0303:


modules/gui/surveillance.lua
```diff
@@ -174,3 +174,34 @@ Event.add(defines.events.on_tick, function(_)
...
+                        current_camera_set.buttons.table[cctv_player.name].selected_index = 1
+                    end
+
+                    local selected_index = current_camera_set.buttons.table[cctv_player.name].items[current_camera_set.buttons.table[cctv_player.name].selected_index]
+                    current_camera_set['cctv_display'].position = game.players[selected_index].position
+                    current_camera_set['cctv_display'].surface_index = game.players[selected_index].surface_index
+
+                else
+                    current_camera_set['cctv_display'].position = {x=0, y=0}
+                    current_camera_set['cctv_display'].surface_index = game.surfaces['nauvis'].index
```
Cooldude2606, line 202: Don't need this code here because a player is not going to notice a one tick delay between the index changing and the next time `on_tick` is called.

Cooldude2606 (changes requested):
Very happy with this PR. Only a single change needed.

Cooldude2606 (approved):

PR #323, opened 2024-08-31
Locale

PHIDIAS0303:


Cooldude2606 (changes requested):
This breaks `/chelp` and `Commands.search`. Also please use `description` not `descr` there is no reason to use a short hand.

Cooldude2606 (changes requested):
This is a breaking change, find a way to extend commands instead of replacing existing functionality. A hot fix way would be to add a `searchable_description` property to commands which contains the orginal english text as a string.

modules/commands/roles.lua
```diff
@@ -12,14 +12,14 @@ local format_chat_player_name, format_chat_colour_localized = _C.format_chat_pla
...
-Commands.new_command('assign-role', 'Assigns a role to a player')
+Commands.new_command('assign-role', {'expcom-roles.description-assign-role'}, 'Assigns a role to a player')
 :add_param('player', false, 'player-role')
 :add_param('role', false, 'role')
 :set_flag('admin-only')
 :add_alias('rpromote', 'assign', 'role', 'add-role')
 :register(function(player, action_player, role)
     local player_highest = Roles.get_player_highest_role(player)
-    if player_highest.index < role.index then
+    if player_highest and (player_highest.index < role.index) then
```
Cooldude2606, line 22: This additional check is not needed, the player always has at least one role so will always have a highest. Revert this and the other ones in this file.

modules/commands/spectate.lua
```diff
@@ -9,25 +9,35 @@ require 'config.expcore.command_general_parse'
...
-Commands.new_command('spectate', 'Toggles spectator mode')
+Commands.new_command('spectate', {'expcom-spectate.description-spectate'}, 'Toggles spectator mode')
 :register(function(player)
     if Spectate.is_spectating(player) then
         Spectate.stop_spectate(player)
+
     else
+        if Spectate.is_following(player) then
+            Spectate.stop_follow(player)
+        end
```
Cooldude2606, line 20: This check and the one below are not needed. The control code already handles this. Remove this addition and the one below.

Cooldude2606 (changes requested):
Please remove the unnecessary checks you added, everything else works as expected.

modules/commands/spectate.lua
```diff
@@ -21,13 +22,18 @@ end)
...
-Commands.new_command('follow', 'Start following a player in spectator')
+Commands.new_command('follow', {'expcom-spectate.description-follow'}, 'Start following a player in spectator')
 :add_alias('f')
 :add_param('player', false, 'player-online')
 :register(function(player, action_player)
     if player == action_player then
        return Commands.error{'expcom-spectate.follow-self'}
+
     else
+        if Spectate.is_spectating(player) then
```
Cooldude2606, line 33: Also need to remove this redundant check.

Cooldude2606 (approved):

PR #327, opened 2024-09-07
Bonus GUI

PHIDIAS0303:
fix combined bonus
add auto personal battery recharge


modules/gui/bonus.lua
```diff
@@ -146,6 +181,8 @@ Gui.element{
         end
     end
 
+    disp[disp['bonus_display_personal_battery_recharge'].tags.counter].caption = format_number(disp['bonus_display_personal_battery_recharge'].slider_value)
+
```
Cooldude2606, line 185: The code you currently have causes and error, and when fixed it doesn't reset the value. So you need the code below.
```suggestion
    local slider = disp['bonus_display_personal_battery_recharge_slider']
    slider.slider_value = config.player_special_bonus['personal_battery_recharge'].value
    disp[slider.tags.counter].caption = format_number(slider.slider_value)

```

config/bonus.lua
```diff
@@ -37,6 +36,18 @@ return {
         ['cpdb'] = 'character_item_pickup_distance_bonus'
         ]]
     },
+    player_special_bonus_rate = 300,
+    player_special_bonus = {
+        ['personal_battery_recharge'] = {
+            -- 1 MW
+            value = 8,
```
Cooldude2606, line 43: This starting value is too high for the base number of points. A value of 2 would work, and although it is not centered, I do not see many people making use of it more than that.

Cooldude2606 (changes requested):

config/bonus.lua
PHIDIAS0303: it is not that high when you are using it to do robots. 4 MW each

Cooldude2606: Then please increase the base number of points, because the default exceeds it.

Cooldude2606 (approved):

PR #331, opened 2024-09-24
Locale

PHIDIAS0303:
add the remaining locale for the command in other directories, such as addons


modules/data/personal-logistic.lua
```diff
@@ -18,7 +20,7 @@ local function pl(type, target, amount)
     end
 
     for _, v in pairs(config.request) do
-        c(config.start + v.key)
```
Cooldude2606, line 21: Please also remove `start` from the config file.

Cooldude2606 (approved):

PR #343, opened 2024-12-30
Warp list

PHIDIAS0303:
the collision mask name changed


Cooldude2606 (approved):
New value confirmed by [factorio data raw](https://wiki.factorio.com/Data.raw#collision-layer)

PR #347, opened 2025-01-07
Decon Log

PHIDIAS0303:
Solve:
Error while running event level::on_player_ammo_inventory_changed (ID 40)
__level__/modules/exp_legacy/modules/addons/deconlog.lua:103: attempt to index field 'character' (a nil value)

Also a change with groupping the series of action as one config


Cooldude2606 (changes requested):
This does not work, it no longer logs explosive rockets or nukes. Also keep all config options.

Cooldude2606 (approved):

PR #348, opened 2025-01-07
Compilatron

PHIDIAS0303:
Error while running event level::on_nth_tick(900)
value for required field 'position' is missing
stack traceback:
	[C]: in function 'create_entity'
	__level__/modules/exp_legacy/modules/addons/compilatron.lua:98: in function 'spawn_compilatron'
	__level__/modules/exp_legacy/modules/addons/compilatron.lua:47: in function 'handler'


Cooldude2606 (changes requested):
I can not accept this solution, the entity must be spawned where it will not collide with something, using a default doesn't not provide this guarantee.

Cooldude2606 (approved):

PR #349, opened 2025-01-08
Module Exports

PHIDIAS0303:
Error while running event level::on_character_corpse_expired (ID 103)
__level__/modules/exp_util/module_exports.lua:423: attempt to index local 'entity' (a nil value)
stack traceback:
    __level__/modules/exp_util/module_exports.lua:423: in function 'get_storage_for_stack'
    __level__/modules/exp_util/module_exports.lua:462: in function 'copy_items_to_surface'
    __level__/modules/exp_util/module_exports.lua:495: in function 'transfer_inventory_to_surface'
    ...vel__/modules/exp_legacy/modules/addons/death-logger.lua:155: in function 'handler'
    __level__/modules/exp_legacy/utils/event.lua:22: in function 'handler'

### DIFF
--- a/exp_util/module/module_exports.lua
+++ b/exp_util/module/module_exports.lua
@@ -420,7 +420,7 @@ function ExpUtil.get_storage_for_stack(options)
     local current, count, entities = cache.current, cache.count, cache.entities
     for i = 1, cache.count do
         local entity = entities[((current + i - 1) % count) + 1]
-        if entity.can_insert(item) then
+        if entity and entity.can_insert(item) then
             cache.current = current + 1
             return entity
         end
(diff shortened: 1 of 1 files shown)


Cooldude2606:
`entity` can not be nil here, there is a deeper issue.

PHIDIAS0303:
i think some user died when cleaning biters. i not very sure what issue.

Cooldude2606 (changes requested):
Until this can be reproduced, the error should not be silently ignored as implemented by this PR. I would suggest raising an error if entity is nil and include more infomation about the context such as the cache contents and the value of i.

Cooldude2606 (approved):

PR #353, opened 2025-01-09
vlayer

PHIDIAS0303:
some change in methods and other


Cooldude2606 (approved):
I will accept this for now because it makes it compatible with 2.0. However, this will allow for infinite solar panel creation by allowing more normal quality solar panels to be removed than were put in, which can then be recycled and made back into higher quality solar panels.

PR #355, opened 2025-01-14
vlayer

PHIDIAS0303:
display surface and the diff of landfill with two mods


Cooldude2606:
Code looks good, but can you give a justification for increasing surface area of landfill when space age is enabled?

Cooldude2606:
I see, the recipe was changed from 20 to 50 in 2.0.7; however it was a base game change not a space age change, so this does not need a dependency check on space age.

Cooldude2606 (approved):

PR #364, opened 2025-01-22
Waterfill

PHIDIAS0303:
added option to allow the use of old days recipe.
as you cannot do that in runtime, so, just by having the ingredient in inventory.


exp_scenario/module/commands/waterfill.lua
```diff
@@ -59,7 +61,15 @@ Selection.on_selection(SelectionName, function(event)
 
     surface.set_tiles(tiles_to_make, true, "abort_on_collision", true, false, player, 0)
     local remaining_tiles = surface.count_tiles_filtered{ area = area, name = "water-mud" }
-    player.remove_item{ name = "cliff-explosives", count = tile_count - remaining_tiles }
+    local t_diff = tile_count - remaining_tiles
+
+    if item_count_cliff >= area_size then
```
Cooldude2606, line 66: This condition is always true because of the check on line 45. You need better logic here to be able to use both recipes.

Cooldude2606:
This feature still does not work as intended. You need to consider the total of both methods when calculating if the player has enough for the area, then find out how much can use the first recipe and now much needs to use the second recipe.

PHIDIAS0303:
^ i did not resolve the conversation

Cooldude2606 (commented):
Logic looks correct. I will test in game once I have availabilty. Then this should be good for merge.

Cooldude2606 (approved):
Your code works fine, however I have found a bug in the unmodified code which I will need to look into.

PR #365, opened 2025-01-22
vlayer space

PHIDIAS0303:
allow power for spaceship. less large ship means better ups.


Cooldude2606:
Power requirements of a space platform are one or the main challenges of their existence, this removes too much of that challenge for simply "better ups". If you want this feature merged you will either need to show significant gains in ups or add a config option to allow it and have the default as false.

Cooldude2606 (approved):

PR #366, opened 2025-01-26
Kill

PHIDIAS0303:
some graphics effect


Cooldude2606 (commented):
Logic is correct, but I will want to test how lighting damange interacts with this before merging.

Cooldude2606 (approved):
I could not find any way for this to cause damange, safe to merge.

PR #368, opened 2025-01-28
non english locale issue

PHIDIAS0303:
it is nil on other language.


Cooldude2606 (approved):
Will approve this merge; however, I will want to look further into why these are nil.

PR #369, opened 2025-01-30
Refactor core library "gui" to new package "exp_gui"

Cooldude2606:
Refactored and upgraded the Gui library in the following ways:
- Add LuaLS type annotations
- Add element tracking and global element iterator
- Add data storage which can be indexed by player, force, or element
- Add shorthand for properties defined directly by arguments
- Change event names to match factorio api
- Unify how draw, style, and data is defined
- Stabilised names to allow for save patching
- Moved toolbar settings into core package
- Complete redesign of top level functions
- Converted all existing GUIs to new libaray


exp_gui/index.ts
```diff
@@ -0,0 +1,8 @@
+import * as lib from "@clusterio/lib";
+
+export const plugin: lib.PluginDeclaration = {
+	name: "exp_gui",
+	title: "exp_gui",
+	description: "Example Description. Plugin. Change me in index.ts",
```
PHIDIAS0303, line 6: example descr :)

Cooldude2606: All the others are the same, I think I will opt to update these all in a single patch rather than change it in this PR.

PR #371, opened 2025-02-03
EXP GUI Tree Decon

PHIDIAS0303:
Error while running event level::on_gui_click (ID 1)
LuaPlayer doesn't contain key player_index.
stack traceback:
    [C]: in function '__index'
    __level__/modules/exp_gui/control.lua:40: in function 'get_player'
    __level__/modules/exp_legacy/modules/addons/tree-decon.lua:47: in function 'handler'
    __level__/modules/exp_gui/prototype.lua:406: in function 'raise_event'
    __level__/modules/exp_gui/prototype.lua:391: in function 'handler'


Cooldude2606 (approved):
Surprised I only missed one of these with how many of them there are.

PR #375, opened 2025-02-06
kill

PHIDIAS0303:
reverse logic


Cooldude2606 (approved):
Needed to remove the `=` as this would allow people of the same role to use the command on each other.

PR #377, opened 2025-02-07
ADDON miner

PHIDIAS0303:
some other way to do


exp_legacy/module/modules/addons/miner.lua
```diff
@@ -161,27 +159,27 @@ Event.add(defines.events.on_resource_depleted, function(event)
         return
     end
 
-    local entities = event.entity.surface.find_entities_filtered{ area = { { event.entity.position.x - 1, event.entity.position.y - 1 }, { event.entity.position.x + 1, event.entity.position.y + 1 } }, type = "mining-drill" }
-
-    if #entities == 0 then
-        return
-    end
+    local resource = event.entity
+    local drills = resource.surface.find_entities_filtered{ type = "mining-drill" }
```
Cooldude2606, line 163: This will check every miner on the surface, we should not be doing this for performance reasons. Re-add the `area` option to the search filter.

Cooldude2606 (approved):

PR #378, opened 2025-02-07
GUI Research

PHIDIAS0303:
some logic change


PHIDIAS0303:
option to allow disable research after some sort of levels. for example, artillery range to prevent the map size being too large

PHIDIAS0303:
moving nuke protect feature out

exp_legacy/module/modules/gui/research.lua
```diff
@@ -64,36 +63,26 @@ end
...
-    if res_n < 3 then
-        res_n = 3
-    elseif res_n > (#research.time - 5) then
-        res_n = #research.time - 5
-    end
-
-    return res_n
+    local max_start = math.max(1, #res.disp - 7)
+    local start = math.clamp(current - 3, 1, max_start)
+    return math.min(start, max_start)
```
Cooldude2606, line 85: `math.min` is not required here because `start` is always lte to `max_start` because of the clamp on the line above.

exp_legacy/module/modules/gui/research.lua
```diff
@@ -105,68 +94,75 @@ local function research_notification(event)
...
+        local rq = event.research.force.research_queue
+
+        for i = #rq, 1, -1 do
+            if rq[i] == event.research.name then
+                table.remove(rq, i)
             end
         end
+
+        event.research.force.cancel_current_research()
+        event.research.force.research_queue = rq
```
Cooldude2606, line 130: This is called after a research is finished, which means it will not stop nukes being researched if the limit is set to 1.

exp_legacy/module/modules/gui/research.lua
```diff
@@ -105,68 +94,75 @@ local function research_notification(event)
...
+            difference = "",
+            color = font_color["positive"]
         }
 
-        local res_i = res_n + i - 3
-
-        if res["disp"][res_i] then
-            res_disp[i]["name"] = { "research.res-name", res["disp"][res_i]["raw_name"], prototypes.technology[res["disp"][res_i]["raw_name"]].localised_name }
+        if entry.raw_name then
+            data.name = { "research.res-name", entry.raw_name, prototypes.technology[entry.raw_name].localised_name }
```
Cooldude2606, line 150: Please pull in the assert I added on main.

Cooldude2606 (changes requested):

Cooldude2606 (approved):

PR #379, opened 2025-02-07
GUI Bonus

PHIDIAS0303:
some changes


exp_legacy/module/modules/gui/bonus.lua
```diff
@@ -365,11 +302,15 @@ Event.add(defines.events.on_player_created, function(event)
...
+    local player = game.players[event.player_index]
+    bonus_score_limit = bonus_score_limit_calc(player)
+    apply_bonus(player)
 end)
 
 Event.add(Roles.events.on_role_unassigned, function(event)
-    apply_bonus(game.players[event.player_index])
+    local player = game.players[event.player_index]
+    bonus_score_limit = bonus_score_limit_calc(player)
+    apply_bonus(player)
```
Cooldude2606, line 313: This will cause desyncs because the value of `bonus_score_limit` is different for each player.

exp_legacy/module/modules/gui/bonus.lua
```diff
@@ -4,13 +4,23 @@
...
+
 local bonus_container
 
+local bonus_data = {
+    score_limit = {}
+}
+
+Storage.register(bonus_data, function(tbl)
+    bonus_data = tbl
+end)
```
Cooldude2606, line 22: The value is only read after a call to `bonus_score_limit_calc` which means there is no reason to store this data. What you are looking for in a memorise function where you can also clear the value when the player's role changes. It might look something like this:
```lua
local _player_bonus_limits = {} -- does not need to be in storage
local function get_player_bonus_limit(player)
  if not _player_bonus_limits[player] then
    _player_bonus_limits[player] = bonus_score_limit_calc(player)
  end
  return _player_bonus_limits[player]
end

local function on_role_assigned(event)
  local player = game.get_player(event.player_index)
  _player_bonus_limits[player] = nil
end
```

exp_legacy/module/modules/gui/bonus.lua
```diff
@@ -365,11 +307,13 @@ Event.add(defines.events.on_player_created, function(event)
 end)
 
 Event.add(Roles.events.on_role_assigned, function(event)
-    apply_bonus(game.players[event.player_index])
```
Cooldude2606, line 368: You can't remove these lines, we still need to update the bonus applied when a player's roles change.

exp_legacy/module/modules/gui/bonus.lua
```diff
@@ -113,58 +121,16 @@ local bonus_gui_control_pts_a = Gui.element("bonus_gui_control_pts_a")
...
-        name = Gui.property_from_name,
-        caption = "0",
-        style = "heading_2_label",
+        caption = "0 / 0",
+        value = 0,
+        style = "electric_satisfaction_statistics_progressbar",
     }:style{
         width = config.gui_display_width["half"],
+        font = "heading-2",
     }
```
Cooldude2606, line 134: Need to add `color` here so that the bar is not green when the limit is exceeded. 
```suggestion
    }:style{
        width = config.gui_display_width["half"],
        font = "heading-2",
        color = { 1, 0, 0 },
    }
```

Cooldude2606 (changes requested):
When a player is demoted, this can reduce the amount of points available to them. Therefore, within `apply_bonus` we need to reset bonus to 0 if the player exceeds their limit.

PR #380, opened 2025-02-12
GUI Production

PHIDIAS0303:


exp_legacy/module/modules/gui/production.lua
```diff
@@ -16,21 +16,38 @@ local precision = {
...
+    local formatted = string.format("%.2f", amount / scale)
+    -- Split into integer and fractional parts
+    local integer_part, fractional_part = formatted:match("^(%-?%d+)%.(%d+)$")
+    integer_part = integer_part or formatted
+    -- Add commas to integer part
+    integer_part = integer_part:reverse():gsub("(%d%d%d)", "%1,"):reverse()
+    integer_part = integer_part:gsub("^,", ""):gsub("-,", "-")
+    -- Handle numbers below 1000 without suffix
+    -- Combine parts and add suffix
+    return string.format("%s.%s%s", integer_part, fractional_part or "00", (suffix == "" and "") or (" " .. suffix)):gsub("%.?0+ %k$", " " .. suffix)
```
Cooldude2606, line 50: Rather than using `" " .. suffix` instead add these spaces in `suffix_list` e.g. `[" G"] = 1e9` this removes most the checks being done here. Also could you explain the purpose of `gsub("%.?0+ %k$", " " .. suffix)` and fyi `%k` is an invalid escape, which is also why I am confussed on the purpose of this call.

Cooldude2606 (changes requested):

Cooldude2606:
I have fixed an issue with the table title sizes. But also added a new tooltip, please can you add your locales then I will merge.

Cooldude2606 (approved):

PR #382, opened 2025-02-24
Control vlayer

PHIDIAS0303:
combinator descr.
logistic filter.
fix delete section will cause issue.

### DIFF
--- a/exp_legacy/module/modules/control/vlayer.lua
+++ b/exp_legacy/module/modules/control/vlayer.lua
@@ -333,6 +333,7 @@ end
 -- @treturn LuaEntity The entity that was created for the interface
 function vlayer.create_input_interface(surface, position, circuit, last_user)
     local interface = surface.create_entity{ name = "storage-chest", position = position, force = "neutral" }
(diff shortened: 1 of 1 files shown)


exp_legacy/module/modules/control/vlayer.lua
```diff
@@ -333,6 +333,7 @@ end
 -- @treturn LuaEntity The entity that was created for the interface
 function vlayer.create_input_interface(surface, position, circuit, last_user)
     local interface = surface.create_entity{ name = "storage-chest", position = position, force = "neutral" }
+    interface.storage_filter = { name = "deconstruction-planner", quality = "normal" }
```
Cooldude2606, line 336: I am not sure why this is needed? The chest is on the neutral force so no logistic bots will interact with it.

PHIDIAS0303: they are for people to read. not written for the bot

Cooldude2606: What are people reading from this?

Cooldude2606 (approved):

PR #385, opened 2025-04-12
Command Research

PHIDIAS0303:
The scenario level caused a non-recoverable error.
Please report this error to the scenario author.

Error while running event level::on_research_finished (ID 21)
__level__/modules/exp_scenario/commands/research.lua:74: attempt to index field '?' (a nil value)
stack traceback:
    __level__/modules/exp_scenario/commands/research.lua:74: in function 'handler'


Cooldude2606 (approved):
Your fix did not address the root of the issue, therefore I have updated your additional checks to use `assert` rather than ignoring the error.

PR #416, opened 2026-03-19
fix inventory clear

PHIDIAS0303:
on_player_banned - player_name
on_player_kicked - player_index

the first event have player_index as optional output.
accepting player_name should be fixing the issue.


Cooldude2606 (changes requested):
This will not fix the issue, the `assert` will still throw an error as the player is not in the game. Instead you need `if not event.player_index thenr return end`

Cooldude2606 (approved):

PR #417, opened 2026-05-02
extra_logging spaces and fix to player_died

PHIDIAS0303:


Cooldude2606:
You should not add spaces to the strings, this is already handled by the concat function within add log line.

Can you please describe the bug you are fixing as I do not see any change in logic by your code.

PHIDIAS0303:
L40

Cooldude2606:
> L40

This is not an explanation. I can see this is where you made an edit but how does checking the type make a difference when we already check if the by player is valid. What bug are you fixing with this additional if statement.

Cooldude2606 (approved):
Removed the redudant check but you were right with this bug after you provided the traceback on discord. Issue was that `.player` will raise an error if the type is not `character`

PR #428, opened 2026-06-24
remove personal-logistic

PHIDIAS0303:
since the user could make the personal logistic from existing logistic / circuit groups, and the function is out of order, it is proposed to be removed.


Cooldude2606:
There might still be use for this, but its been broken for so long that it has gone out of use. So im happy to merge this.

PR #429, opened 2026-06-24
bonus changes

PHIDIAS0303:
dropped support for the legacy command, use the gui instead.
the initial value of each bonus will be the half of its max value.
reduced bonus scale.


Cooldude2606:
Happy with these changes, but merge conflict needs resolved.

Issue #425, opened 2026-06-23
auto research


PHIDIAS0303:
### Details
Since the circuit control for lab is introduced, the auto research function should be obsolete.
Features should be dropped to reduce complexity.

PR #430, opened 2026-06-24
research changes

PHIDIAS0303:
Also some other changes to the module.
Removed bonus inventory from research.
#425


Cooldude2606:
Auto research can be removed, but I think the inventory bonus should remain.

Cooldude2606 (approved):

PR #433, opened 2026-06-24
remove landfill blueprint

PHIDIAS0303:
This is obsolete with the super force build.


Cooldude2606:
This never worked perfectly and I agree it is no longer needed. So happy to merge.

Issue #424, opened 2026-06-23
locale suggestions


PHIDIAS0303:
### Details
Some links should be in a config file instead of inside a locale file.
Some features, such as selection area, should share the same locale as well.

PR #434, opened 2026-06-24
lawnmower changed to selection

PHIDIAS0303:
and added to quick action
locale need change with #424


exp_scenario/module/commands/lawnmower.lua
```diff
@@ -2,32 +2,59 @@
...
-        local surface = player.surface
-        
-        -- Intentionally left as player.position to allow use in remote view
-        local entities = surface.find_entities_filtered{ position = player.position, radius = range, type = "corpse" }
-        for _, entity in pairs(entities) do
-            if (entity.name ~= "transport-caution-corpse" and entity.name ~= "invisible-transport-caution-corpse") then
-                entity.destroy()
-            end
+local Selection = require("modules/exp_util/selection")
+local SelectArea = Selection.connect("ExpCommand_Waterfill")
```
bbassie: This reuses waterfill's selection channel — `waterfill.lua:9` is also `Selection.connect("ExpCommand_Waterfill")`. In `selection.lua`, handlers live in a shared `_registered[name]` table and `event_dispatch` looks them up by `active_selection.name`, so lawnmower and waterfill end up sharing handlers and selection state: selecting an area for one also fires the other's `on_selection`, and `:stop()` (which compares `active_selection.name ~= self.name`) cross-cancels between them. Use a unique name, e.g. `Selection.connect("ExpCommand_Lawnmower")` (which also matches the `ExpCommand_Lawnmower` annotation above).

exp_scenario/module/commands/lawnmower.lua
```diff
@@ -2,32 +2,59 @@
...
+    local surface = event.surface
+    local area_size = (area.right_bottom.x - area.left_top.x) * (area.right_bottom.y - area.left_top.y)
+
+    if area_size > 1000 then
+        player.print({ "exp-commands_waterfill.area-too-large", 1000, area_size }, Commands.print_settings.error)
+        return
+    end
+
+    -- Intentionally left as player.position to allow use in remote view
+    local entities = surface.find_entities_filtered{ position = player.position, area = area, type = "corpse" }
```
bbassie: `position` has precedence over `area` in `find_entities_filtered`/`find_tiles_filtered`/`destroy_decoratives`, and with no `radius` it matches only that single point — so the selected `area` is ignored and only the player's current tile is cleared. The old radius-based code needed `position`, but for an area selection it should be dropped; compare `waterfill.lua:68`, which passes `area = area` with no `position`. Same fix applies on lines 48 and 54.

exp_scenario/module/commands/lawnmower.lua
```diff
@@ -2,32 +2,59 @@
...
+        replace_tiles[i] = { name = "grass-1", position = tile.position }
+    end
+
+    surface.set_tiles(replace_tiles)
+    surface.destroy_decoratives{ position = player.position, area = area }
 
-        surface.set_tiles(replace_tiles)
-        surface.destroy_decoratives{ position = player.position, radius = range }
-    end)
+    player.print({ "exp-commands_waterfill.complete", tile_count }, Commands.print_settings.default)
```
bbassie: `tile_count` is never declared in this handler (it resolves to a `nil` global), unlike `waterfill.lua:65` which has `local tile_count = 0`. The completion message will interpolate nil — use `#replace_tiles` (or a real counter).

exp_scenario/module/commands/lawnmower.lua
```diff
@@ -2,32 +2,59 @@
...
+--- @class ExpCommand_Lawnmower.commands
+local commands = {}
+
+--- Toggle player selection mode for lawnmower
+--- @class ExpCommands_Lawnmower.commands.lawnmower: ExpCommand
+--- @overload fun(player: LuaPlayer)
+commands.lawnmower = Commands.new("lawnmower", { "exp-commands_lawnmower.description" })
+    :register(function(player)
+        if SelectArea:stop(player) then
+            return Commands.status.success{ "exp-commands_waterfill.exit" }
```
bbassie: These user-facing strings use the `exp-commands_waterfill.*` namespace (also lines 24, 35, and 56), so the lawnmower command prints waterfill's text. I see the description notes locale changes are pending via #424 — flagging so these move to `exp-commands_lawnmower.*` before merge.

bbassie (commented):
Thanks for porting lawnmower to the selection workflow — the structure mirrors `waterfill` nicely. I focused on correctness and found a few copy-paste carryovers from `waterfill` worth addressing before merge (details inline).

Summary:
1. **Selection channel name collides with waterfill** — entangles the two commands' handlers and selection state.
2. **`position = player.position` left alongside `area`** — makes the selected area ignored, so only the player's tile is cleared.
3. **`tile_count` is undefined** in the completion message.
4. **Locale keys still point at `exp-commands_waterfill.*`** — acknowledged via #424.

Items 1–3 are not covered by the pending-locale note and will make the command misbehave or silently no-op. (Reviewing as a comment — I only have read access to this repo.)

Cooldude2606 (approved):

PR #435, opened 2026-06-25
clear blueprint changed to selection

PHIDIAS0303:
the 3 function into quick action


bbassie (commented):
Nice rework — folding these into quick actions and using a drag-selection for the bounded clear is a clear improvement, and the selection logic mirrors the existing `waterfill` command correctly (name-scoped dispatch, `event.surface`, `AABB.expand`, the 1000-tile cap all match). The `ExpCommand_ClearBlueprint` selection name doesn't collide with the others, and every role reference to the old command names is fully renamed. A few things to address before merge:

### 1. The three new quick actions have no caption/tooltip locale (breaks the buttons)
`new_quick_action(name, …)` builds each button from `exp-gui_quick-actions.caption-<name>` / `tooltip-<name>`, but the `en.cfg` change only touched `[exp-commands_surface]`. There are no `caption-clear-ground-item`, `caption-clear-blueprint-surface`, `caption-clear-blueprint` (or their tooltips), so all three buttons render as `Unknown key: "exp-gui_quick-actions.caption-clear-ground-item"` for every player. The `[exp-gui_quick-actions]` section needs the six new keys.

### 2. `zh-CN.cfg` / `zh-TW.cfg` `[exp-commands_surface]` not updated
The keys were renamed in `en.cfg` (`description-items` → `description-item`, plus new `item`/`blueprint`/`enter`/`exit`/`area-too-large`/`complete`), but the zh files still hold the old keys and lack the new ones. Chinese-locale players will see `Unknown key` for the command descriptions and every result message (and the new quick-action labels).

### 3. `complete=__1__ tiles were handled.` reports a ghost count, not tiles
The selection handler clears `entity-ghost`s and prints `#entities`, but the message says "tiles". (The "tiles" wording is correct in the `waterfill` command it was copied from, since that one sets tiles — here it's the number of ghosts cleared.) e.g. clearing 5 ghosts prints "5 tiles were handled."

### 4. Lost attribution + inconsistent audience between the new commands
`clear_ground_item` and `clear_blueprint_surface` use `game.print{ … }` with generic, no-name messages ("Cleared the items found on the current surface."), whereas the old commands named the acting player and surface. Meanwhile `clear_blueprint` uses `player.print` (actor only). So a surface-wide clear is broadcast to everyone with no record of who ran it or where, while the selection clear is silent to others. Worth keeping player/surface attribution and making the audience consistent for these shared admin actions.

### 5. `clear-blueprint-surface` and `clear-blueprint` share the same description
Both register with `{ "exp-commands_surface.description-blueprint" }` ("Clear all blueprints."), so they're indistinguishable in `/help` and the quick-action tooltips despite one being surface-wide and the other a bounded selection.

#1 is the one that visibly breaks the feature for all players, and #2 for the zh locales; #3–#5 are polish.

Cooldude2606 (changes requested):
This is a regression in the behaviour of `/clear-ground-item` and `/clear-blueprinsts` as it is no longer possible to select a different surface. The surface argument must remain. I think it is fine to change the default from "whole game" to "player surface" but you must use the `:defaults` pattern as seen in other commands such as `/set-always-day`

Cooldude2606 (approved):

PR #437, opened 2026-06-25
repair add quick action, changed to selection

PHIDIAS0303:
The hardcoded list is change to using prototype hidden.
This will allow more item to be on the list.
They are not supposed to be acquired from the normal game, so no actual experience harmed.
If you can use them, you are using editor mode anyway.

### DIFF
--- a/exp_legacy/module/config/repair.lua
+++ b/exp_legacy/module/config/repair.lua
@@ -2,7 +2,6 @@
 -- @config Repair
 
 return {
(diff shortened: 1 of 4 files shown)


Cooldude2606:
This PR does more than the title suggests. I think range should remain on the command, and please can you describe the impact of changing to prototype hidden over using the hard coded list.

Cooldude2606 (approved):

PR #446, opened 2026-06-27
remove limit

PHIDIAS0303:


Cooldude2606:
This isnt an increase, this is a removal. And if you are going to remove it then you need to also removed the unused size variable.

Cooldude2606 (approved):

PR #448, opened 2026-07-31
Add exp_roles plugin and scenario permissions

bbassie:
Replaces the role storage half of `exp_legacy/expcore/roles.lua` with clusterio's own roles and permissions. Clusterio already stores roles, the permissions they grant, and which user holds which role, plus a web UI for all three — so this adds only what it is missing, rather than reimplementing any of it.

This does not yet move the call sites off `expcore.roles`; nothing depends on the new plugin. That is deliberate, see "What is left" below.

## `exp_scenario`: permissions for the in game actions

The role config held 105 action strings and 7 flags that only existed as bare strings in lua. Each is now a real clusterio permission, so it appears on the role page and can be granted like any other.

Names come from a deterministic transform, exported so the lua side derives exactly the same name and existing call sites keep working untouched:

```
command/kill                   -> exp_scenario.command.kill
gui/warp-list/bypass-proximity -> exp_scenario.gui.warp_list.bypass_proximity
fast-tree-decon                -> exp_scenario.action.fast_tree_decon
report-immune          (flag)  -> exp_scenario.flag.report_immune
```

Flags become permissions too — `player_has_flag` and `player_allowed` are the same lookup, the only difference being that flags have change triggers, which stay in lua.

Command descriptions are taken from the existing locale entries rather than written fresh, so they match what players already see in `/commands`. Two permissions were referenced in code but missing from the role config (`command/give-warning`, `command/collectdata`); both are now defined.

The 21 actions the Guest role granted are marked `grantByDefault`, since the in game default role and clusterio's default role are the same concept.

Three actions have **no** scenario permission — they map onto core ones instead:

| Legacy action | Replaced by |
|---|---|
| `command/assign-role` | `core.user.update_roles` |
| `command/unassign-role` | `core.user.update_roles` |
| `command/get-roles` | `core.role.list` |

## `exp_roles`: the sync plugin

`RoleUpdatesEvent` and `UserUpdatesEvent` are both `dst: "control"`, so instances cannot subscribe to them. The controller plugin rebroadcasts on its own events, the same shape `exp_groups` uses.

It stores one record per role for the properties clusterio's role does not carry — order, priority, short hand, tag, colour, auto assign threshold, block auto assign — created automatically for any role without one and removed when the role is deleted. These are edited on the core role page through a `componentExtra`, so there is no second place to manage roles.

There is no `default_role_id` or `root_role_id` setting: the default role is `controller.default_role_id`, and root is `core.admin`, which already short circuits every permission check.

`exp_roles.sync_mode` is `disabled` / `enabled` / `bidirectional`, matching `exp_groups`.

### Priority replaces `disallow`

Only the roles with the highest priority a player holds are considered, and those union as normal. Jail sits above everything else, so holding it suppresses every other role including the implicit default — which is what `disallow(default.allowed)` was reaching for. It never worked, as the field is `allowed_actions`, not `allowed`, so it was passing nil.

This means jail lives on the cluster like any other role, and `Jail.old_roles` is no longer needed: jailing becomes "add one role", unjailing "remove one role", with no stashing and restoring.

### Local assignments

Assignments made in game are applied locally first and then sent up, so `assign_player` stays synchronous for callers and nothing downstream has to learn about the round trip. Once the controller confirms, the role moves out of the local overlay. If it is refused, the instance rolls the change back.

The same overlay carries assignments which should never leave the map — `assign_player_local`, for roles earned from time on this map.

### Auto assign

Roles earned from online time across the cluster are granted by the controller from the threshold on the role, using the online time clusterio already tracks. Per map conditions stay in lua as local assignments. This drops the legacy `MapsPlayed` / `AfkTime` conditions in favour of online time alone.

## Verification

No factorio server was involved, so the lua was exercised directly under a stubbed environment:

- The transform was compared against the TypeScript across all 114 names (104 actions, 3 core mapped, 7 flags) — no mismatches. This is the one invariant that silently breaks every permission check if it drifts.
- 28 behavioural assertions over the real module covering priority suppression, the implicit default role, the local overlay, pending confirmation, rejection rollback, the server bypass, ordering, and the legacy `config.players` view.
- Message records round trip through their own typebox schemas.
- Types, the web bundle, and lua parsing all clean.

## What is left

Follow ups, kept separate because they are the risky part and deserve their own review:

- Move the ~40 files off `require("modules.exp_legacy.expcore.roles")`, simplify jail, delete the legacy module and config.
- Seed the existing roles on first run.

`override_player_roles` is not migrated, and there is no automatic export of live save data — `Storage.register` keys on `debug.getinfo` short_src and asserts it is not at runtime, so an rcon helper would register under the wrong name. A manual `_export` taking the table as an argument can be added if it turns out to be wanted.

## Also

One unrelated one line fix: `exp-commands_unjail.description` was a copy of the `/jail` text.

### DIFF
--- a/exp_roles/controller.ts
+++ b/exp_roles/controller.ts
@@ -1,0 +1,299 @@
+import { BaseControllerPlugin, InstanceRecord } from "@clusterio/controller";
+import * as lib from "@clusterio/lib";
+import * as messages from "./messages";
+import * as path from "node:path";
+
+export class ControllerPlugin extends BaseControllerPlugin {
+	roleMeta!: lib.SubscribableDatastore<messages.RoleMetaRecord>;
+
+	async init() {
+		const databaseDirectory = this.controller.config.get("controller.database_directory");
+
+		this.roleMeta = new lib.SubscribableDatastore(
+			...await new lib.JsonIdDatastoreProvider(
+				path.join(databaseDirectory, "exp_roles", "role_meta.json"),
+				messages.RoleMetaRecord.fromJSON.bind(messages.RoleMetaRecord),
+			).bootstrap()
+		);
+
+		// Roles created before this plugin was installed have no properties yet
+		this.ensureRoleMeta();
+
+		this.controller.subscriptions.handle(messages.RoleUpdatedEvent, this.handleRoleSubscription.bind(this));
+		this.controller.subscriptions.handle(
+			messages.AssignmentUpdatedEvent, this.handleAssignmentSubscription.bind(this)
+		);
+
+		this.roleMeta.on("update", this.roleMetaUpdated.bind(this));
+		this.controller.roles.on("update", this.rolesUpdated.bind(this));
+		this.controller.users.records.on("update", this.usersUpdated.bind(this));
+
+		this.controller.handle(messages.RoleListRequest, this.handleRoleListRequest.bind(this));
+		this.controller.handle(messages.RoleMetaUpdateRequest, this.handleRoleMetaUpdateRequest.bind(this));
+
+		this.controller.handle(messages.AssignmentListRequest, this.handleAssignmentListRequest.bind(this));
+		this.controller.handle(messages.AssignmentUpdateRequest, this.handleAssignmentUpdateRequest.bind(this));
+	}
+
+	async onShutdown() {
+		await this.roleMeta.save();
+	}
+
+	/*
+		Role properties
+	*/
+
+	/** Create properties for any role which does not have them yet, ordered after every existing role. */
+	ensureRoleMeta() {
+		const created = [];
+		let nextOrder = 0;
+		for (const meta of this.roleMeta.values()) {
+			nextOrder = Math.max(nextOrder, meta.order);
+		}
+
+		for (const role of this.controller.roles.values()) {
+			if (!this.roleMeta.has(role.id)) {
+				nextOrder += 1;
+				created.push(new messages.RoleMetaRecord(role.id, nextOrder));
+			}
+		}
+
+		if (created.length) {
+			this.roleMeta.setMany(created);
+		}
+
+		return created;
+	}
+
+	/** Combine a clusterio role with its in game properties. */
+	buildRoleRecord(role: Readonly<lib.Role>) {
+		const meta = this.roleMeta.get(role.id) ?? new messages.RoleMetaRecord(role.id, role.id);
+		return new messages.RoleRecord(
+			role.id,
+			role.name,
+			[...role.permissions],
+			meta,
+			role.id === this.controller.config.get("controller.default_role_id"),
+			Math.max(role.updatedAtMs, meta.updatedAtMs),
+			role.isDeleted,
+		);
+	}
+
+	async onControllerConfigFieldChanged(field: string, curr: unknown, prev: unknown) {
+		// Which role is the default is carried on the role records themselves
+		if (field === "controller.default_role_id") {
+			this.controller.subscriptions.broadcast(new messages.RoleUpdatedEvent(this.listRoleRecords()));
+		}
+	}
+
+	/** Every role which currently exists, in the form sent to instances. */
+	listRoleRecords() {
+		return [...this.controller.roles.values()].map(role => this.buildRoleRecord(role));
+	}
+
+	async handleRoleListRequest() {
+		return this.listRoleRecords();
+	}
+
+	async handleRoleMetaUpdateRequest(request: messages.RoleMetaUpdateRequest) {
+		const meta = request.meta;
+		if (!this.controller.roles.has(meta.id)) {
+			throw new lib.RequestError(`Role with ID ${meta.id} does not exist`);
+		}
+
+		this.roleMeta.set(meta);
+	}
+
+	/** A clusterio role was created, changed or deleted. */
+	rolesUpdated(roles: lib.Role[]) {
+		const created = this.ensureRoleMeta();
+
+		// Deleting a role leaves its properties behind, which would be reused by
+		// the next role to be given the same id
+		const orphaned = [];
+		for (const role of roles) {
+			if (role.isDeleted) {
+				const meta = this.roleMeta.getMutable(role.id);
+				if (meta) {
+					orphaned.push(meta);
+				}
+			}
+		}
+		if (orphaned.length) {
+			this.roleMeta.deleteMany(orphaned);
+		}
+
+		// Newly created properties broadcast on their own through roleMetaUpdated
+		const createdIds = new Set(created.map(meta => meta.id));
+		const updates = roles.filter(role => !createdIds.has(role.id)).map(role => this.buildRoleRecord(role));
+		if (updates.length) {
+			this.controller.subscriptions.broadcast(new messages.RoleUpdatedEvent(updates));
+		}
+
+		this.applyAutoAssign();
+	}
+
+	/** The in game properties of a role changed. */
+	roleMetaUpdated(metas: messages.RoleMetaRecord[]) {
+		const updates = [];
+		for (const meta of metas) {
+			const role = this.controller.roles.get(meta.id);
+			// A deleted role broadcasts through rolesUpdated instead
+			if (role) {
+				updates.push(this.buildRoleRecord(role));
+			}
+		}
+
+		if (updates.length) {
+			this.controller.subscriptions.broadcast(new messages.RoleUpdatedEvent(updates));
+		}
+
+		this.applyAutoAssign();
+	}
+
+	async handleRoleSubscription(request: lib.SubscriptionRequest) {
+		const roles = this.listRoleRecords().filter(role => role.updatedAtMs > request.lastRequestTimeMs);
+		return roles.length ? new messages.RoleUpdatedEvent(roles) : null;
+	}
+
+	/*
+		Assignments
+	*/
+
+	/**
+	 * The roles of a user as seen in game.
+	 *
+	 * The default role is left out because every player has it, which keeps the
+	 * assignments to only those users who have been given a role.
+	 */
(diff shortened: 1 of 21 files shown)


bbassie:
On the red lint check: it is pre-existing, not from this branch. `main` at a07be7b7 — the commit this branches from — already fails it ([run](https://github.com/explosivegaming/ExpCluster/actions/runs/29294225887)), and the job fails on any finding at all.

The repo currently reports 279 findings across 51 files. This branch contributes **zero** — I ran the same LuaLS 3.18.2 and FMTK factorio library the workflow uses and fixed all 17 the new module initially had.

Happy to leave it, or to fix the pre-existing ones in a separate PR if that would be useful — `exp_scenario/module/commands/home.lua` (47) and `exp_commands/module/module_exports.lua` (30) account for a large share of them.

exp_roles/index.ts
```diff
@@ -0,0 +1,77 @@
...
+	title: "List Role Assignments",
+	description: "List the roles held by each player in game.",
+	grantByDefault: true,
+});
+lib.definePermission({
+	name: "exp_roles.assignment.subscribe",
+	title: "Subscribe to Role Assignment Updates",
+	description: "Receive updates when the roles held by a player change.",
+	grantByDefault: true,
+});
```
Cooldude2606: We should not use new custom permissions here. Instead we should be using `core.roles`

`exp_roles.assignment` is not required as we do not need to make assignments visible to control. This is an implimentation detail for allowing instances to interact with the core role and user models.

exp_roles/controller.ts
```diff
@@ -0,0 +1,299 @@
...
+
+		this.roleMeta = new lib.SubscribableDatastore(
+			...await new lib.JsonIdDatastoreProvider(
+				path.join(databaseDirectory, "exp_roles", "role_meta.json"),
+				messages.RoleMetaRecord.fromJSON.bind(messages.RoleMetaRecord),
+			).bootstrap()
+		);
+
+		// Roles created before this plugin was installed have no properties yet
+		this.ensureRoleMeta();
```
Cooldude2606: We also need to ensure there are no orpahns as the datastore file might have exist from a previous install of the plugin. In effect we are running all of `rolesUpdated` on init so it might be useful to have `ensureRoleMeta` `sweepRoleMeta` and `applyAutoAssign` which are all called by `rolesUpdated` and `init`

exp_roles/controller.ts
```diff
@@ -0,0 +1,299 @@
...
+			Math.max(role.updatedAtMs, meta.updatedAtMs),
+			role.isDeleted,
+		);
+	}
+
+	async onControllerConfigFieldChanged(field: string, curr: unknown, prev: unknown) {
+		// Which role is the default is carried on the role records themselves
+		if (field === "controller.default_role_id") {
+			this.controller.subscriptions.broadcast(new messages.RoleUpdatedEvent(this.listRoleRecords()));
+		}
```
Cooldude2606: We should use a switch statement here like we do in instance.ts

exp_roles/module/control.lua
```diff
@@ -0,0 +1,901 @@
...
+    script_data.synced_players = {}
+    for _, record in pairs(payload.assignments) do
+        if not record.is_deleted and record.role_ids then
+            script_data.synced_players[record.name] = record.role_ids
+        end
+    end
+
+    -- Anything still pending was sent before the connection was lost, the
+    -- controller state received here is authoritative
+    script_data.pending = {}
```
Cooldude2606: Don't we also need to clean up `local_players` as some of the pending assignments may now be confirmed and no longer need to be held within `local_players` as `synced_players` now holds it.

exp_roles/module/control.lua
```diff
@@ -0,0 +1,901 @@
...
+        local role = script_data.roles[role_id]
+        if role then roles[#roles + 1] = role end
+    end
+
+    if #roles == 0 then return end
+
+    local emit_updates = script_data.emit_updates
+    script_data.emit_updates = false
+    change_player_roles(player_name, roles, "unassign", "<server>", true, false)
+    script_data.emit_updates = emit_updates
```
Cooldude2606: Do we need to mutate `emit_updates` here? The `sync` param is false which means we never check `emit_updates`

exp_roles/instance.ts
```diff
@@ -0,0 +1,104 @@
...
+
+		const [roles, assignments] = await Promise.all([
+			this.instance.sendTo("controller", new messages.RoleListRequest()),
+			this.instance.sendTo("controller", new messages.AssignmentListRequest()),
+		]);
+
+		await this.luaSend("initialise", {
+			roles: roles.map(role => role.toJSON()),
+			assignments: assignments.map(assignment => assignment.toJSON()),
+		});
```
Cooldude2606, line 54: If possible, can we benchmark this command. We have recently found that commands containing a large number of string literals can harm performance. In our case it would be the permission strings. Two alternatives would be a permission array to convert an index to string value allowing reuse between roles; another would be a nested structure e.g. `{ "core": { "roles": [ "list", "subscribe" ] } }`. The goal is better time performance. Include your results in your reply and any chalanges you faced in the implimention, commiting the best one. Feel free to explore other alternatives to stay away from custom binary formats that require base64 encoding. If the gains are minimal, then keep this simpler method.

exp_roles/web/components/RoleProperties.tsx
```diff
@@ -0,0 +1,145 @@
...
+			color,
+			values.autoAssignHours === null || values.autoAssignHours === undefined
+				? null
+				: values.autoAssignHours * MS_PER_HOUR,
+			values.blockAutoAssign ?? false,
+		)));
+	}
+
+	return <>
+		<SectionHeader title="In Game Properties" />
```
Cooldude2606: Can we arrange this is a columed format, or other alterntive that is nicer presented. Currently this takes up a lot of room and it is not clear what the apply button does or that it is specific to the in game properties. Also the "block auto assign" should have a tooltip explaining that it blocks other roles from being autoassigned.

Cooldude2606 (changes requested):

bbassie:
Pushed dcba1eef covering all seven points. Details on the two that needed judgement.

## Benchmark of the initialise command

Measured against a real headless Factorio 2.1.12 over RCON, decoders loaded as a scenario so only the payload travels in the command. Test data is the current role config: 15 roles, 847 permission entries across them, 109 unique, plus a 250 player roster. All three encodings were checksummed first and produce byte-identical results (`847:24372`).

**The dominant cost is not the payload — it is the tick.** An empty `/sc bench.noop()` round trip is 16.65 ms, and 60 commands pipelined back to back took 1000 ms whether they carried 0 bytes or 38 KiB. Factorio processes one RCON command per tick, so a single initialise is free at any of these sizes.

Isolating parse+decode by looping 500 times inside one command:

| encoding | size | parse+decode | vs flat |
|---|---|---|---|
| flat (current) | 37.5 KiB | 0.50 ms | — |
| **indexed** | **17.2 KiB** | **0.37 ms** | **26% faster** |
| nested | 29.5 KiB | 0.95 ms | 90% slower |

Nested loses because reconstructing each permission string needs concatenation during the tree walk, which costs more than it saves on transfer.

Stressing to 4x the roles, where the gap matters more:

| encoding | size | parse+decode |
|---|---|---|
| flat | 122.1 KiB | 1.52 ms |
| **indexed** | **30.2 KiB** | **0.87 ms** |

**Committed the indexed form.** On time alone the gain is minimal at current scale — 0.13 ms once per instance start, under 1% of a tick — so by your "keep the simpler method" rule flat would have been fine. What tipped it is that the win compounds: at 4x roles it is 43% faster and a quarter the size, because flat repeats every permission string once per role while indexed sends each unique name once. 122 KiB in a single command is the part I would rather not grow into. It cost about 15 lines, and single role updates stay in the plain form since they have nothing to share.

Two things worth flagging:

- **The nested form has a correctness trap.** `exp_scenario.command.kill` and `exp_scenario.command.kill.always` both exist, so a node is simultaneously a leaf and a branch. The obvious encoding silently drops one. I needed a `"*"` terminal marker, which also eats into the size saving. Worth knowing if it comes up again.
- I could not check your string-literal concern the way I first tried, because globals set by one `/sc` do not survive into the next — each command is sandboxed. I moved the decoders into a scenario. I did test the underlying worry directly: a command built as a Lua table constructor with 847 string literals (38.8 KiB) versus one `helpers.json_to_table` blob measured identically, both inside tick noise. So the current approach was never exposed to that problem.

## Permissions

Dropped all five custom permissions. Roles now use `core.role.list` / `core.role.subscribe` / `core.role.update`, and the web UI checks `core.role.update`. Assignments dropped `control` from their routing entirely and carry no permission, since as you say they are an implementation detail between controller and instance. `AssignmentUpdateRequest` keeps `core.user.update_roles` because it does mutate the core user model.

## The rest

`ensureRoleMeta` / `sweepRoleMeta` / `applyAutoAssign` are now separate and all three run on init as well as from `rolesUpdated`. The sweep is a real fix, not just tidying: properties left by a role deleted while the plugin was off would previously have been inherited by whatever role next got that id.

On the initialise cleanup you spotted, the two cases differ, so I kept them distinct. An incremental update only drops roles the controller has confirmed, since a confirmation may still be in flight. `initialise` is authoritative, so anything still pending is either confirmed and now in `synced_players`, or it never landed and has to be given up. Both stop being held locally; `assign_player_local` roles are untouched either way. Added four assertions covering it, 36 total, all passing, and LuaLS still reports zero findings for the module.

bbassie:
Done in 1943d874 — added `labelAlign="left"`, which was defaulting to `"right"` and pushing each label up against its input.

Layout is otherwise unchanged, so the columns and tooltips stay as reviewed.

Cooldude2606 (approved):

PR #450, opened 2026-08-06
Migrate the lua lint from luals to emmylua

bbassie:
Supersedes #449, which pinned `factoriomod-debug` to 2.1.3 as a stopgap. fmtk is unpinned again here and tracks latest. Now also carries what was briefly split into #451 and #452, folded back in so this is one review.

The check ends at **0 findings across 187 files**, with all but six diagnostics enabled.

## Why

fmtk 2.1.4 onwards retargets its generated typedefs at emmylua and luals misreads them: 2.1.4 switched dicts to `table<K, V>`, and 2.1.5 redefined `LocalisedString` as `[string, LocalisedString...]`, which luals 3.18.2 cannot match against a plain `string`. On the pre-fix tree that is 0 findings with fmtk 2.1.3, 11 with 2.1.4 and 279 with 2.1.8. Bumping luals is not an option, 3.18.2 is already the latest release, and every fmtk changelog since 2.1.5 says luals support "will be dropped completely in a near future version, and may already be broken in some areas".

Measured the other way round, the newer typedefs are clearly better for emmylua: against the same source the 2.1.3 library gives 226 findings and 2.1.8 gives 0.

## The migration

`.emmyrc.json` replaces `.luarc.json`. The `--clusterio-modules` plugin has no emmylua equivalent, as emmylua has no plugin interface, so it is reimplemented as `workspace.moduleMap`. Those rules rewrite each file's own module path rather than the require string, which is why they read backwards. `.luarc.json` is now git ignored so the factoriomod-debug extension can write machine local library paths there; emmylua merges both files.

The workflow installs a prebuilt `emmylua_check` (7 MB, no rust toolchain) and takes a sparse checkout of `clusterio/clusterio` for `modules/clusterio/*`. Library paths are merged in through a second `-c` config rather than the two `jq` passes over the committed config. The gate counts diagnostics rather than trusting the exit code, which only reflects error severity, and asserts that at least 100 files were read so the check cannot pass by analysing nothing.

`type.patch.lua` keeps its role as the place for api corrections, but its old contents are gone: `raise_event` now takes `LuaEventType` in the generated typedefs and `LuaObject` has become `LuaObject.base`. `exp_scenario/.luacheckrc` is deleted, luacheck is never invoked.

## Patching the position types

Per review, `MapPosition` and `BoundingBox` are patched back to their named form:

```lua
---@alias MapPosition MapPosition.struct
---@alias BoundingBox BoundingBox.struct
```

The api documents both as a union with a positional array because both are accepted as input, which makes every read of `.x` or `.left_top` optional and accounted for a third of the 340 `need-check-nil` findings. Everything read back from the game uses the named form, and writing it that way everywhere is the better convention, so the positional variant is removed rather than cast around at each call site.

Two places were still writing the positional form and are converted: six offsets in the spawn area config, and the two search areas in mine depletion. `apply_offset` no longer needs its `position.x or position[1]` fallback.

## Two things that were silently doing nothing

emmylua deserialises an unrecognised diagnostic code to a catch all that matches no checker, so every `invisible`, `nil-check` and `global-element` suppression in the repo was inert. They are renamed to the emmylua codes.

More significantly, an inline cast is only parsed as `--[[@as T]]`. All 149 casts in the repo were written `--[[ @as T ]]` and were being treated as plain comments. Normalising them is called out in CONTRIBUTING.md, since it is easy to reintroduce.

## Bugs the checker found

- `config.temp_warning_limit` does not exist. The warnings config defines `script_warning_limit`, and the locale expects it as `__3__/__4__`, so `/warnings` was printing nil.
- `graftorio/forcestats.lua` iterated `get_contents()` and `targeted_items_pickup`/`_deliver` as name keyed dictionaries. Factorio 2.0 made them `ItemWithQualityCount[]`, so the logistorage stats were keying off array positions.
- `LogisticFilter.signal` became `value` in 2.0, so the vlayer loop that clears stale circuit signals broke on the first slot and left outdated values behind.
- `Gui.add_top_element` and friends assert on `Gui.top_elements[define.name]`, but the tables are keyed by the define, so the duplicate registration assert never fired.
- `GuiIter.remove_element` indexed `registered_scopes` by player index instead of the scope elements it had just fetched, so it removed nothing.
- `rocket_info.lua`'s prune loop shadowed its own parameter, so `silo_data[unit_number] = nil` wrote into the entry instead of removing it.
- `assert` returns every argument it is given, so `return assert(x, "message")` was returning the message as a second value from nine functions annotated with one.
- `Async.AsyncFunction` declared `@operator call` with no parameters, so every async call looked over supplied.
- `loadstring` was removed in Lua 5.2, so the debug model's `dump_text` would have errored.
- `trains.lua` declared its command as `ExpCommand_Artillery.commands.artillery`, and `Async`'s `id`, `func_id` and `next_id` were documented as numbers when `get_function_name` returns a string.

## Behaviour changes

- `Roles._prototype:set_permission_group` returned nil for an unknown group, which broke the config chain with `attempt to index a nil value` several lines later. It now asserts with the group name.
- `spawn_area` no longer accepts the positional offset form, which is why the config entries changed with it.
- `get_tile` is documented as taking `x` and `y`; four calls passed a position instead.

## Still disabled

`unnecessary-if` (152) is disabled by the config fmtk generates for itself and its findings are all "condition is always truthy", which the api union types make unreliable. `unused` (170) is mostly the fixed gui handler signatures, 145 of them are `def`, `player`, `element` and `event`. `unnecessary-assert` (35) and `preferred-local-alias` (18) are style. `assign-type-mismatch` (99) and `param-type-mismatch` (191) are worth a later pass; their original justification was a luals performance problem, "time reduced to 10s from 400s", which no longer applies since the whole check now runs in under a second.

## Verifying

From the repo root, with the fmtk library generated by `fmtk luals-addon` and a config adding it plus clusterio's `packages/host` to `workspace.library`:

```
emmylua_check . -c .emmyrc.json,ci.emmyrc.json -f json --output check.json
```

A clean run was checked against a canary file to confirm it still catches a bad `LuaPlayer` field, a bad ExpCluster module member and an unresolvable require, and `unresolved-require`, `type-not-found`, `undefined-global` and `syntax-error` are all 0, so the pass is not vacuous.


Cooldude2606 (changes requested):
I am happy with all the lua code; however, the CI needs to be fixed as it currently does not use the factorio types correctly.

Cooldude2606 (approved):

PR #451, opened 2026-08-07
Enable need-check-nil

bbassie:
Stacked on #450, which is where the emmylua migration itself lives. That PR left `need-check-nil` disabled with a note; this one clears the backlog and removes it from the disable list, so the check now runs with nil safety on at **0 findings across 187 files**.

## Patching the position types

Per review, `MapPosition` and `BoundingBox` are patched back to their named form in `type.patch.lua`:

```lua
---@alias MapPosition MapPosition.struct
---@alias BoundingBox BoundingBox.struct
```

The api documents both as a union with a positional array because both are accepted as input, which makes every read of `.x` or `.left_top` optional and accounted for a third of the 340 findings. Everything read back from the game uses the named form, and writing it that way everywhere is the better convention, so the positional variant is removed rather than cast around at each call site. That alone took 340 down to 30.

Two places were still writing the positional form and are converted: six offsets in the spawn area config, and the two search areas in mine depletion. `apply_offset` no longer needs its `position.x or position[1]` fallback.

## Everything else

The rest is per site. Where a value is genuinely always present the invariant is asserted, which also gives a clear message rather than an index error on nil:

- `Roles.get_player_highest_role` never returns nil, `get_player_roles` always returns at least the default or root role.
- `Roles.new_role` returned the result of `error`, which made its return nullable.
- Element data lookups, row lookups and `debug.getinfo` frames are asserted where the entry is created alongside the thing being looked up.
- Forward declared locals assigned during control setup are given types rather than being inferred as nil.
- `LuaControl.force` is a `ForceID` union, so force only members need the cast the repo already uses elsewhere.

Types were tightened where the annotation was simply wrong: the task list element data was a union of three table types, which made every lookup optional, and is now one class with an index signature per key type. The legacy role record gained a class, and the vlayer storage tables gained element types.

## Behaviour changes

Two are deliberate and worth flagging:

- `Roles._prototype:set_permission_group` returned nil for an unknown group, which broke the config chain with `attempt to index a nil value` several lines later. It now asserts with the group name.
- `spawn_area` no longer accepts the positional offset form, which is why the config entries changed with it.

## Verifying

Same as #450. A clean run was checked against a canary file to confirm it still catches a bad `LuaPlayer` field, a bad ExpCluster module member and an unresolvable require, and `unresolved-require`, `type-not-found`, `undefined-global` and `syntax-error` are all 0.


bbassie:
Folded into #450 so the whole change is one review rather than three cumulative diffs. Nothing is lost, the commits are now on that branch.

PR #452, opened 2026-08-07
Enable the remaining practical diagnostics

bbassie:
Stacked on #451. Enables ten more diagnostics and clears the 51 findings they reported, so the check still runs at **0 findings across 187 files**.

## What is now enabled

Five reported nothing at all and were only disabled because the luals config demoted them: `await-in-sync`, `missing-return-value`, `unbalanced-assignments`, `undefined-doc-param` and `invert-if`. They were masking nothing except a future regression.

Five more are correctness checks worth the 51 findings they cost: `redefined-local`, `redundant-return-value`, `missing-parameter`, `redundant-parameter` and `return-type-mismatch`.

## What that found

`assert` returns every argument it is given, so `return assert(x, "message")` was returning the message as a second value. Nine functions annotated with a single return value were handing back two.

`get_tile` is documented as taking `x` and `y`. Four calls passed a position instead, which is what the old `Incorrect Api Type` suppressions were about.

`Async.AsyncFunction` declared `@operator call` with no parameters, so every async call looked over supplied. It now uses the `@overload` form the gui element classes already use.

The status helpers in exp_commands built their message with `msg == nil and default or msg`, which types as optional. They now test explicitly, which also preserves the case where a localised string is the boolean `false`. `internal_error` took a required message while `Commands.Status` declares it optional.

Four locals shadowed an outer one of the same name. The report command's broadcast loop shadowed the reporting player, and research_milestones declared `force` twice in the same scope, the second only to add a cast.

`Elements.centered_label` is called with no caption in the readme to make spacers, so the caption is now optional in its overload rather than the call being wrong.

## Cleaning up after #451

Enabling `unnecessary-assert` locally showed that 15 of the asserts added during the nil check burn down were redundant once the surrounding annotations landed, so they are removed here. `unnecessary-assert` itself stays disabled, it still reports 35 pre-existing cases.

It also caught two real annotation gaps: `_has_handlers` and `_track_elements` were inferred as the literal `false` from the prototype table, which made `assert(self._has_handlers, ...)` statically impossible. Both are now declared `boolean`.

## Left disabled

`unnecessary-if` (152) is disabled by the config fmtk generates for itself and its findings are all "condition is always truthy", which the api union types make unreliable. `unused` (170) is mostly the fixed gui handler signatures, 145 of them are `def`, `player`, `element` and `event`. `assign-type-mismatch` (99) and `param-type-mismatch` (191) are worth a later pass. Their original justification was a luals performance problem, "time reduced to 10s from 400s", which no longer applies since the whole check now runs in under a second.


bbassie:
Folded into #450 so the whole change is one review rather than three cumulative diffs. Nothing is lost, the commits are now on that branch.

PR #453, opened 2026-08-19
Move the scenario onto exp_roles and remove the legacy role system

bbassie:
Part two of #448. Moves every call site off `expcore.roles`, deletes the legacy module and config, simplifies jail, and seeds the roles on first run. Taking "improve the role api with no restraint, no external consumers other than the mono repo" at its word, the lua api is redesigned rather than kept compatible, and the permission names are tidied up while nothing holds them yet.

## Lua api

The legacy action strings and the transform which mapped them onto permission names are gone. Call sites check the clusterio permission directly, so `grep exp_scenario.command.kill` finds the definition, the check, and the seed, and there is no second transform which silently breaks every check if it drifts.

| Before | After |
|---|---|
| `player_allowed(p, "command/kill")` | `player_has_permission(p, "exp_scenario.command.kill")` |
| `player_has_flag(p, "report-immune")` | `player_has_permission(p, "exp_scenario.bypass.reports")` |
| `define_flag_trigger(flag, fn)` | `define_permission_trigger(permission, fn)` — any permission can have one |
| `events.on_role_assigned` / `on_role_unassigned` | `events.on_player_roles_changed` with `assigned` and `unassigned` names |
| `get_player_highest_role(a).index < get_player_highest_role(b).index` | `player_outranks(a, b)`, plus `player_outranks_role(a, role)` |
| `get_role_from_any` / `get_role_by_name` / `get_roles_ordered` | `get_role(name \| id \| role)` / `get_roles()` |
| `config.order` / `config.roles` / `config.players` | removed; `role:get_player_names()` covers the one use |
| `assign_player(p, roles, by, skip_checks, silent)` | `assign_player(p, roles, by, silent)` |

Flags were only permissions with a change trigger, so the separate concept goes. Every consumer registered both role events for the same handler, so they become one, and it is now also raised when the roles a player holds change on the controller, or when a role they hold is edited there. The old events were never raised for either, which left guis stale after a web ui change. `player_outranks` applies the core.admin bypass consistently; two of the six index comparisons did not.

A role now carries a **permission group**, which the legacy system mapped roles to and #448 dropped. A player is moved into the group of their most privileged role which names one, which is how Jail actually restricts a player. It sits with the other in game properties on the role page.

## Permission names

Explicit list in `exp_scenario/permissions.ts`, no transform. The `action` and `flag` buckets only reflected how the old config was written, so: `exp_scenario.bypass.*` (entity protection, nuke protection, deconstruction log, reports), `exp_scenario.decon.*` (standard, fast trees — with descriptions that say what they gate), `exp_scenario.player.*` (admin, spectator, instant respawn, system commands), `exp_scenario.chat.commands`, and `exp_scenario.gui.player_list.kick` / `.ban` for the player list buttons which were never commands. `command.*` and `gui.*` are unchanged apart from `clear_tag.always` → `tag_clear.always` to match the command it belongs to.

Commands derive `exp_scenario.command.<name>` in the authority, so `assign-role`, `unassign-role` and `get-roles` get scenario permissions rather than the core ones #448 mapped them to. The in game command is bounded by the lower role check, `core.user.update_roles` is not — granting it to Moderator would have let any mod with a web account change anyone's roles.

Dropped: `defer_role_changes` (priority replaced it), `command/give-warning` (no role held it; the player list now keys the warn button on `create_warning`, the permission the command already needed) and `command/report` (never defined; now `create_report`). Both buttons were only ever visible to root. `_ipc` and `_sudo` are added so every command has a definition.

## Jail

`jail_player` gives the Jail role, `unjail_player` takes it away. `Jail.old_roles` and the stash/restore are gone, since priority suppresses the other roles for as long as the role is held.

## Seeding

On the first run (empty role properties datastore) the controller creates the roles the scenario shipped with and gives the listed players their roles. The seed keeps the `parent` relationships from the old config and flattens them, since clusterio roles do not inherit; the flattened sets were diffed against the legacy config and match for every role. Cluster Admin and the default role already exist, so those entries only set properties — the default role gets its permissions through `grantByDefault`, which is checked to equal the old Guest list. Roles which already exist by name are reused and only gain permissions, so seeding a cluster which already has some is safe.

## Verification

- 118 behavioural assertions over the real module under a stubbed environment: lookup, priority suppression, ranking, holders, every sync entry point, pending confirmation, rejection rollback, local only roles, events and messages for both local and controller originated changes, triggers and permission groups, jail.
- End to end in a dev cluster: seed on controller start (15 roles, 38 users), instance on Factorio 2.1.14 starts with the full scenario, permission checks over rcon resolve as expected, in game assignment confirmed by the controller, assignment for an unknown user rejected and rolled back, `ctl user set-roles` reaching the game.
- emmylua lint as ci runs it: 0 findings. tsc and the web bundle clean.

## Notes

- `exp_scenario.command.kill.always` and `spawn.always` stay defined; `/kill` never checked `kill.always` before either, its lower role parser already bounds it. Left for a later tidy.
- The seed data is the old config, staff list included, so nothing new is in the repo.
- Web ui screenshots are not included — the `controller run --dev` web build is not set up here, the form only gains one text field next to the existing ones.


exp_roles/module/control.lua
```diff
@@ -61,60 +65,28 @@ ExpRoles._prototype = Role
...
-    permission = core_action_permissions[action]
-    if not permission then
-        local body = action:gsub("%-", "_"):gsub("/", ".")
-        if not body:find(".", 1, true) then
-            body = "action." .. body
-        end
-        permission = "exp_scenario." .. body
+--- Move a player into a permission group, done async because the game does not
+--- allow it from within every event
+local set_permission_group_async = Async.register(function(player, group)
```
Cooldude2606: Roles should not interact with permission groups anymore. This is the responsibility of exp_groups. Therefore this can be removed.

exp_roles/module/control.lua
```diff
@@ -61,60 +65,28 @@ ExpRoles._prototype = Role
 --- @field emit_updates boolean
 local script_data = {}
 
---[[
-    Permission names
-]]
+--- Roles in order, the most privileged first, rebuilt from script data
+local ordered_roles = {} --- @type ExpRoles.Role[]
+--- Roles indexed by name, rebuilt from script data
+local roles_by_name = {} --- @type table<string, ExpRoles.Role>
```
Cooldude2606: We should have scripts use the role id not the name. So we dont need this mapping. When we need to get a role by name (which will be rare) we can search the list of roles.

exp_roles/module/control.lua
```diff
@@ -61,60 +65,28 @@ ExpRoles._prototype = Role
 --- @field emit_updates boolean
 local script_data = {}
 
---[[
-    Permission names
-]]
+--- Roles in order, the most privileged first, rebuilt from script data
+local ordered_roles = {} --- @type ExpRoles.Role[]
```
Cooldude2606: Why do we need to maintain this order? The only time we check index is when doing direct comparisons, so we dont need this list.

exp_legacy/module/config/afk_kick.lua
```diff
@@ -8,6 +8,7 @@ return {
     trust_time = 3600 * 60 * 10, --- @setting trust_time The time in ticks that a player must be online for to count as trusted
     update_time = 3600 * 30, --- @setting update_time How often in ticks the script checks for active players
     custom_active_check = function(player)
-        return Roles.get_player_highest_role(player).index <= Roles.get_role_from_any("Veteran").index
+        local veteran = Roles.get_role("Veteran")
+        return veteran ~= nil and Roles.get_player_highest_role(player).index <= veteran.index
```
Cooldude2606: We should keep "index" internal and instead expose a method for doing comparisons.

exp_roles/module/control.lua
```diff
@@ -171,269 +137,258 @@ local function decode_role(record, names)
...
 
     return false
 end
 
---- Check if a player bypasses all permission checks
---- This replaces the legacy root role
+--- Check if a player is more privileged than a role
+--- A player with core.admin, which includes the server, outranks every role
 --- @param player LuaPlayer | string | nil
+--- @param role string | number | ExpRoles.Role
```
Cooldude2606: It is bad practice to have mutliple ways to specify a role. We should only use role objects only, and have them be methods instead.

exp_roles/module/control.lua
```diff
@@ -171,269 +137,258 @@ local function decode_role(record, names)
...
             return true
         end
     end
 
     return false
 end
 
---- Check if a player holds a role
+--- Check if a role applies to a player
 --- @param player LuaPlayer | string | nil
```
Cooldude2606: We should only accept player objects and not player names.

exp_roles/module/control.lua
```diff
@@ -455,113 +410,158 @@ end
 --- @param message LocalisedString
 function ExpRoles.print_to_roles(roles, message)
```
Cooldude2606: Print should be made into methods on role objects. The lib should then have 'get_higger_roles(role)' and 'get_lower_roles(role)'

exp_roles/module/control.lua
```diff
@@ -678,60 +677,56 @@ end
...
 --- Give a player one or more roles, the change is sent to the controller
 --- @param player LuaPlayer | string
 --- @param roles any A role, role name, or array of either
---- @param by_player_name string?
---- @param skip_checks boolean? Unused, kept for compatibility
---- @param silent boolean?
-function ExpRoles.assign_player(player, roles, by_player_name, skip_checks, silent)
+--- @param by_player_name string? Shown in the game message, defaults to the current player or the server
+--- @param silent boolean? When true no game message is printed
+function ExpRoles.assign_player(player, roles, by_player_name, silent)
```
Cooldude2606: We have two ways to assign roles, one via the lib and one via the role method. We should only have one.

exp_legacy/module/config/gui/player_list_actions.lua
```diff
@@ -23,11 +23,7 @@ end
 
 -- auth that will only allow when on player's of lower roles
 local function auth_lower_role(player, selected_player_name)
-    local player_highest = Roles.get_player_highest_role(player)
-    local action_player_highest = Roles.get_player_highest_role(selected_player_name)
-    if player_highest.index < action_player_highest.index then
-        return true
-    end
+    return Roles.player_outranks(player, selected_player_name)
```
Cooldude2606: We can use this directory rather than as a proxy local function.

exp_roles/web/components/RoleProperties.tsx
```diff
@@ -15,6 +15,7 @@ type RoleFormValues = {
 	shortHand: string;
 	tag: string;
 	color: string | { toRgb(): { r: number, g: number, b: number } } | null;
+	permissionGroup: string;
```
Cooldude2606: As previous comment. We should not have permission groups as part of roles.

exp_roles/controller.ts
```diff
@@ -16,6 +17,12 @@ export class ControllerPlugin extends BaseControllerPlugin {
 			).bootstrap()
 		);
 
+		// An empty datastore means the plugin has not run before, so the roles
+		// the scenario used to define are created on the controller
+		if (this.roleMeta.size === 0) {
```
Cooldude2606: This should be a manual button trigger and not automatic. Only the roles and permissions should be seeded, not the player assignments.

exp_roles/seed.ts
```diff
@@ -0,0 +1,334 @@
...
+	/** Uses the controller's default role rather than creating one. */
+	isDefault?: boolean;
+	/** Uses the controller's admin role rather than creating one. */
+	isAdmin?: boolean;
+	/** Name of the role whose permissions are also granted, applied recursively. */
+	parent?: string;
+	permissions: string[];
+}
+
+const admin = ["exp_scenario.player.admin", "exp_scenario.player.spectator", "exp_scenario.bypass.reports"];
```
Cooldude2606: Can we inline these at the lowest place in the hierarchy. Including these individually is unneeded repition.

exp_scenario/module/commands/_authorities.lua
```diff
@@ -1,26 +1,28 @@
...
+        local permission = "exp_scenario.command." .. command.name:gsub("%-", "_")
+        if not player_has_permission(player, permission) then
             return deny{ "exp-commands-authorities_role.deny" }
         else
             return allow()
         end
     end)
 
-Roles.define_flag_trigger("is_system", function(player, state)
+Roles.define_permission_trigger("exp_scenario.player.system_commands", function(player, state)
```
Cooldude2606: We can change this to be based on core.admin

Cooldude2606 (changes requested):
Along with all the changes highlighted in this review; can you also weigh the pros and cons of using oop for the roles vs only using module level methods. At the moment there are a few places where we have mutliple ways to do things or where things can be cumbersome, so we should align on a strong dx api interface for working with roles and covering the use cases we have or would expect.

bbassie:
Pushed bea8d46c covering every point. The api question first, since it decided the shape of the rest.

## Object methods vs module functions

I weighed it as: module functions are easiest when the role is incidental to the call — a permission check where the caller has a player and a permission name and never thinks about roles, or anything that spans roles such as ordering. Methods are easiest when the caller already holds a role, because there is only one way to say it and the role cannot be given three ways (name, id, object) as it could before. The old api mixed both, which is where the duplication came from: `assign_player` and `assign_player_local`, `player_has_role` next to `role:get_players`, `get_role_from_any` accepting anything.

So the split is now by what the caller has in hand:

- **Module functions take a player or nothing** and answer questions about players or the role list: `get_role(id)`, `get_role_by_name(name)`, `get_roles()`, `get_default_role()`, `get_higher_roles(role)`, `get_lower_roles(role)`, `get_player_roles(player)`, `get_player_highest_role(player)`, `player_has_permission(player, permission)`, `player_outranks(player, other)`, `define_permission_trigger(permission, fn)`.
- **Everything done to or with a role is a method**: `role:assign(player, options)`, `role:unassign(player, options)`, `role:has_player(player)`, `role:has_permission(permission)`, `role:is_higher_than(other)`, `role:is_lower_than(other)`, `role:get_players(online)`, `role:get_player_names()`, `role:print(message)`.

There is one assignment entry point, `role:assign(player, { by_player_name, silent, local_only })`, with local only as an option rather than a second function. Players are `LuaPlayer` only, nil or index 0 for the server. Roles are looked up by id; `get_role_by_name` is a search over the list and is used by the few places which only know a name — the configs, jail, and the role parser — which I think is the right trade rather than maps to keep in step. The name map, the ordered list and the `index` field are gone; `get_roles()` sorts on demand and comparisons go through the methods, `order` stays on the object as data but is documented as not for comparing.

The one cost of methods worth naming: a call site that wants the roles of a player and then something about each of them now does two steps, `get_player_roles` then a method, where a flat `player_has_role(player, "Jail")` was one. Jail is the only place that pattern existed and it reads fine as `jail_role():has_player(player)`.

## The rest

- **Permission groups** removed from roles, the web form and the seed; exp_groups owns that mapping. Lua no longer touches `game.permissions`.
- **Seed** is a `SeedRolesRequest` (`core.role.create`) behind a "Seed roles" button in a section on the roles page, roles and permissions only, no assignments. Existing roles are matched by name and only gain permissions, so it is safe to press again. Verified against the dev cluster by deleting a role and seeding: it came back, the rest were untouched. The seed lists each permission once at the lowest role which has it (`player.admin`, `player.spectator` and `bypass.reports` on Trainee, `instant_respawn` on Moderator and Sponsor) and the parent chain carries it up; the flattened sets still match the legacy config for every role.
- **System commands** unlock on `core.admin`, and `exp_scenario.player.system_commands` is gone. Note this means Senior Administrator can no longer use `/_rcon` unless they also hold Cluster Admin, which the old `is_system` flag allowed — I take that as intended.
- **Player list** auth uses `Roles.player_outranks` directly. The `get_higher_roles` callers (reports, protected entities) loop with `role:print`.
- The event now carries role ids in `assigned` / `unassigned` rather than names, to match.
- While there I registered the role metatable with `Storage.register_metatable`: the role records live in storage and would have lost their methods on load.

Tests updated for the new api, 108 assertions passing; emmylua still 0 findings; tsc and the web bundle clean; the instance starts with the full scenario and the methods answer as expected over rcon.

exp_roles/module/control.lua
```diff
@@ -1,57 +1,68 @@
...
+--- @field unassigned number[] Ids of the roles which were unassigned
+
 --- Methods shared by every role, kept apart from the fields so that defining
 --- them does not count as injecting fields into the role itself
 --- @class ExpRoles.RolePrototype
 local Role = {}
 ExpRoles._prototype = Role
 
+--- Registered so roles keep their methods across save and load
+local role_metatable = Storage.register_metatable("Role", { __index = Role })
```
Cooldude2606: To avoid importing storage we can unroll this and also avoid the module name call by using the plugin name.

exp_roles/module/control.lua
```diff
@@ -61,379 +72,332 @@ ExpRoles._prototype = Role
...
         block_auto_assign = meta.block_auto_assign or false,
-    }, { __index = ExpRoles._prototype })
+    }, role_metatable)
 end
 
---- Refresh the legacy view of the roles a player holds
+--- Sort roles in place, the most privileged first
+--- @param roles ExpRoles.Role[]
+--- @return ExpRoles.Role[]
+local function sort_roles(roles)
```
Cooldude2606: In a number of places we return sorted roles when there is little reason to. This is an extra cost we can remove. In the places where we need it sorted we can expose an ordered roles method or can optimise further in other ways.

exp_roles/module/control.lua
```diff
@@ -61,379 +72,332 @@ ExpRoles._prototype = Role
...
     end
-    for _, role_id in pairs(script_data.local_players[player_name] or {}) do
-        role_ids[#role_ids + 1] = role_id
+    return role_ids
+end
+
+--- Role ids a player has been given, as a set
+--- @param player_name string
+--- @return table<number, true>
+local function get_held_role_set(player_name)
```
Cooldude2606: This is just `seen` from held role ids. Becuase this is an internal method let's just return array, set from the one method as we get both for the same cost.

exp_roles/module/control.lua
```diff
@@ -61,379 +72,332 @@ ExpRoles._prototype = Role
...
+--- Get the names of every player who has been given this role
+--- This includes players who have never joined this map, and is not affected
+--- by priority, so a jailed moderator is still listed under moderator
+--- @param self ExpRoles.Role
+--- @return string[]
+function Role.get_player_names(self)
+    local names, seen = {}, {}
+    for _, players in pairs{ script_data.synced_players, script_data.local_players } do
+        for player_name, role_ids in pairs(players) do
+            if not seen[player_name] then
```
Cooldude2606: This doesnt work becuase a player can be in both lists.

exp_roles/module/control.lua
```diff
@@ -1,57 +1,68 @@
...
-Replaces the legacy role system with clusterio's roles and permissions.
+Mirrors clusterio's roles and permissions into the game.
 
-Roles and the players who hold them are owned by the controller. This module
-mirrors that state into the game and presents the same interface the legacy
-`expcore.roles` module did, so existing call sites keep working.
+Roles, the permissions they grant, and the players who hold them are owned by
+the controller. This module keeps a copy of that state and answers permission
+checks from it, so a check such as `Roles.player_has_permission(player,
+"exp_scenario.command.kill")` is answered from the same data the web ui shows.
```
Cooldude2606, line 7: Although not used yet, let's add "has any" and "has all" the same has how the webui has it.

Cooldude2606 (changes requested):
You mention some tests you are running, do you think those could be added to a test dir within this plugin? We are planning to move away from mocha so can you write them using tap. Please attempt to keep all tests independent, although instance reuse is allowed so long as a method is used fetch / start the instance on its first use rather than having a magic global process always running.

bbassie:
Pushed 8a15745c and e79822b2 covering all six points.

## The api changes

- `player_has_any_permission` and `player_has_all_permission`, varargs like the web ui's account checks, with the same edge semantics: any of none is false, all of none is true.
- Nothing sorts on the way out anymore. `get_ordered_roles()` and a public `sort_roles(roles)` cover the three places which present roles in order (the player list gui, `/get-roles`, the readme welcome line); the highest role is found with a single scan; `get_higher_roles` / `get_lower_roles` and the per-check `get_player_roles` return unsorted.
- `get_held_role_ids` returns the list and the set it was already building, and the second function is gone.
- `get_player_names` collects into a set before listing. You were right to stop on it — the old guard did dedup the both-lists case, but only if you read the placement of `seen` very carefully, and the test suite now pins the case: a role held in both the synced and the local list (local only assign, then the controller assigns the same role) counts once.
- Metatable registered directly as `script.register_metatable("exp_roles.Role", ...)`, storage import dropped.

## Tests

`exp_roles/test/`, tap as the runner: `pnpm --filter @expcluster/roles test` (needs the workspace built first for the seed and message tests, which read from dist). 1092 assertions.

The lua module is tested for real rather than reimplemented: `test/helpers/lua.js` creates a lua state on first use per test file (fengari, so no lua install needed), stubs the factorio surface and the three clusterio modules, and loads `module/control.lua` from disk. Each of the five lua files — lookup, players, assignment, sync, holders — gets its own state, so they are fully independent; the environment is passed to each chunk as an argument, no globals, which also keeps the lint clean over the test files. The javascript side covers the message records round tripping through their typebox schemas, the indexed permission encoding, and the seed: every permission it grants must be defined by exp_scenario, and every parent must exist with its permissions carried into its children.

Writing them caught one real thing beyond the `get_player_names` case: my helper originally returned `assert(role, message)` as a tail call, and assert returns all its arguments, so the message leaked as an extra value in table constructors. The module itself already guarded against this in `get_player_highest_role`, worth knowing as a pattern.

These don't run in this repo's ci since the workspace `@clusterio/*` deps only resolve inside the parent monorepo — same reason there is no js ci today. If you want them wired up once that changes, happy to.

Lint 0 findings, tsc and the web bundle clean, all tests passing.

exp_roles/module/control.lua
```diff
@@ -61,379 +72,378 @@ ExpRoles._prototype = Role
...
-    local names, seen = {}, {}
+    local roles, highest_priority = {}, nil
     for _, role_id in pairs(role_ids) do
         local role = script_data.roles[role_id]
-        if role and not seen[role.name] then
-            seen[role.name] = true
-            names[#names + 1] = role.name
+        if role then
+            roles[#roles + 1] = role
+            if highest_priority == nil or role.priority > highest_priority then
```
Cooldude2606, line 177: Rather than have two loops, we can discard the role array when a new highest is found, and insert into the existing array only if it is equal to the highest.

exp_roles/test/lua/env.lua
```diff
@@ -0,0 +1,193 @@
...
+        env.events, env.printed, env.sent, env.sounds = {}, {}, {}, {}
+    end
+
+    --- Record one test result
+    function env.check(ok, name, detail)
+        env.results[#env.results + 1] = { name = name, ok = not not ok, detail = detail }
+    end
+
+    --- Shallow array equality
+    function env.eq(a, b)
```
Cooldude2606: Can we also add a deep eq.

exp_roles/test/lua/env.lua
```diff
@@ -0,0 +1,193 @@
...
+        { id = 9, name = "Gone", permissions = {}, meta = meta(6), is_deleted = true },
+    }
+end
+
+function Env.assignment(name, role_ids)
+    return { name = name, role_ids = role_ids }
+end
+
+--- Create an independent environment with a fresh copy of the roles module
+function Env.new()
```
Cooldude2606: For all the stubs we impliment, can we add a index metamethod which will raise an error for unknown / unimplimented properties. This more closely mirrors the game api and will help catch errors in tests.

exp_roles/test/messages.test.js
```diff
@@ -0,0 +1,43 @@
+"use strict";
+const t = require("tap");
+const { Value } = require("@sinclair/typebox/value");
+const messages = require("../dist/node/messages");
+
+function roundTrip(subtest, Record, record) {
```
Cooldude2606: Can we also add test matrix similar to what exists for clusterio so that we can have complete coverage of round trips.

Cooldude2606 (changes requested):
For the lua side of the test harness can we do the following two things:
- Add a `env.test` so we can name specific tests and pass env as a param to the test function, allowing a reset each time.
- Make generic stubs and move the file into a folder in the root of the mono repo to allow reuse between plugins. There should still allow a custom env.lua to load additional stubs fpr tests.
For the js tests, can we have coverage of the various plugins parts such as controller and instance. We do not need coverage of the webui.

bbassie:
Pushed 59a0e0be and 24d36db6 covering all five points.

## Harness

The generic parts now live in `test/` at the repository root, ready for other plugins: `test/lua/stubs.lua` (the factorio surface and the clusterio module stubs, recording what the module does to them), `test/lua/framework.lua` (the test registry and check helpers), `test/lua/runner.js` (the fengari runner), and `test/common.js`, which is clusterio's `testMatrix` / `testRoundTripJsonSerialisable` pair. A plugin composes them from its own `test/lua/env.lua` — exp_roles' one adds its fixtures and loads its module, and extends the stub require table, so another plugin only needs its own env.lua.

Tests are declared as `Test.test(name, function(env) ... end)` and every test function receives a fresh environment, so nothing carries between them; a test which errors is reported as a failed check with the error as detail rather than taking the file down. Each stub raises on properties it does not implement — `game.player` is the one read allowed while unset, since the module checks it for the acting player. `Test.deep_eq` added, keys checked from both sides.

## Coverage

- `messages.test.js` runs every record, event and request through the same matrix helpers as the clusterio tests, covering every optional field combination — 256 cases for the meta record alone.
- `controller.test.js` runs the real ControllerPlugin against faked controller internals (a memory datastore for the roles, fake users): property creation and orphan sweeping on init, record building and the default flag, broadcasts from role and property changes, subscription replay filtering, assignment validation and application, auto assignment with the blocking role and the on-leave trigger, and seeding twice to show reuse.
- `instance.test.js` runs the real InstancePlugin against a faked instance: the subscription and initialise flow with the indexed permission encoding asserted on the wire, the missing default role warning, sync mode gating in all three modes, forwarding of both update events, and the rejection rollback reaching lua.

The web ui is not covered, as requested.

## Also

`get_effective_roles` builds its list in one pass — the list restarts when a higher priority is found and appends on a tie.

1147 assertions passing, lint 0 findings, tsc clean.

test/lua/framework.lua
```diff
@@ -0,0 +1,109 @@
...
+            if not Test.deep_eq(value, b[key]) then return false end
+        end
+        for key in pairs(b) do
+            if a[key] == nil then return false end
+        end
+        return true
+    end
+
+    --- Names of an array of roles
+    function Test.names(roles)
```
Cooldude2606: This should not be in the shared test lib as it is role specific.

test/lua/stubs.lua
```diff
@@ -0,0 +1,118 @@
+--[[-- Generic factorio stubs for plugin module tests
+Provides the surface a clusterio module touches, recording what the module does
+to it. Every stub raises an error when an unimplemented property is read, which
+mirrors the game api and catches mistakes in tests early.
+
+Plugins compose these from their own test/lua/env.lua, adding stubs of their
+own through `stubs.requires` and `Stubs.strict`.
```
Cooldude2606: I would prefer an "extends" type method to extend on the requires table rather than directly mutstimg it. It does a deep merge, so if a module already exists then the new methods are added to it, if no module exists then it is added, this is recursive. If a module already exists then an error is raised. Similar should exist for the globals script, game, and defines.

test/lua/stubs.lua
```diff
@@ -0,0 +1,118 @@
...
+function Stubs.new()
+    local stubs = {
+        events = {},  -- events raised through script.raise_event
+        printed = {}, -- game.print and player.print messages
+        sent = {},    -- payloads sent through the clusterio api
+        sounds = {},  -- sounds played to players
+    }
+
+    local next_event_id = 100
+    script = Stubs.strict("script", {
```
Cooldude2606: Should be LuaBootstrap

test/lua/stubs.lua
```diff
@@ -0,0 +1,118 @@
...
+
+    defines = Stubs.strict("defines", {
+        events = Stubs.strict("defines.events", {
+            on_player_joined_game = 1,
+            on_multiplayer_init = 2,
+        }),
+    })
+
+    local players_by_name, players_by_index, connected = {}, {}, {}
+    game = Stubs.strict("game", {
```
Cooldude2606: Should be LuaGameScript

test/lua/stubs.lua
```diff
@@ -0,0 +1,118 @@
...
+            end,
+        })
+        players_by_name[name] = player
+        players_by_index[index] = player
+        if player.connected then connected[#connected + 1] = player end
+        return player
+    end
+
+    --- A player object which represents the server
+    stubs.server = Stubs.strict("server player", { index = 0, name = "<server>" })
```
Cooldude2606: Should be LuaPlayer <server>

test/lua/stubs.lua
```diff
@@ -0,0 +1,118 @@
...
+
+    --- Modules resolved by the stubbed require, extend before loading the module
+    stubs.requires = {
+        ["modules/clusterio/api"] = Stubs.strict("clusterio api", {
+            send_json = function(channel, data)
+                stubs.sent[#stubs.sent + 1] = { channel = channel, data = data }
+            end,
+        }),
+        ["modules/clusterio/compat"] = Stubs.strict("clusterio compat", { script_data = {} }),
```
Cooldude2606, line 139: How do we test that script data is processed correctly with script.on_load? We need have a way to populate script data for a test, and also have register metatable apply correctly.

test/lua/framework.lua
```diff
@@ -0,0 +1,109 @@
...
+The runner creates a fresh environment for every test, so tests are
+independent and can not leak state into each other.
+]]
+
+local Framework = {}
+
+--- Create a test registry, one is made per test file
+function Framework.new()
+    --- @class Test
+    local Test = {
```
Cooldude2606: Can this be Suite since it is a collection of tests.

exp_roles/test/lua/env.lua
```diff
@@ -0,0 +1,84 @@
...
+    return env
+end
+
+--- Run the declared tests, each against a fresh environment
+function Env.finish()
+    Env.Test.run(Env.new)
+    return Env.Test.results_json()
+end
+
+return Env
```
Cooldude2606: Can we add a comment explaing thst the return value becomes `...` in the test files. Also I feel that Env is the wrong name for this because you dont "finish" an env. Its also easily confused with Test, Suite, and Framework. I think we need better defined names between these parts and have better interaction between thrm. Although I quite happy with the number of parts namely stubs/harness -> module specifc extension -> tests.

test/lua/framework.lua
```diff
@@ -0,0 +1,109 @@
+--[[-- Test framework for plugin module tests
+Test files declare named tests with `Test.test(name, function(env) ... end)`.
+The runner creates a fresh environment for every test, so tests are
+independent and can not leak state into each other.
+]]
+
+local Framework = {}
+
+--- Create a test registry, one is made per test file
+function Framework.new()
```
Cooldude2606: Can we have a general `pass` and `fail` method added, which check, eq, and deep_eq all call.

Cooldude2606 (changes requested):
The test framework is very close now, still some refinement needed to get a good dx. As for the tests themselves can we structure them similar to clusterio core where we name them after the files / classes / methods (for both lua and ts) this makes it easy to check we have coverage and understand where new tests should be added. This of course only works for unit tests / stubbed tests; integration tests should be process based similar to what has currently been written.

bbassie:
Pushed 7adcdde9 covering all seven points.

## Naming between the parts

The three parts are now: **Stubs** (the shared factorio surface), the plugin's **env.lua** (the module specific extension, building environments and owning the fixtures), and a **Suite** of tests. The framework's registry is the suite, and it is created around the environment factory — `Framework.new_suite(make_env)` — so env.lua ends with exactly that call and the interaction between the parts is one line. The suite returned there becomes `...` in each test file, and both env.lua and the test files now say so where it happens. "finish" is gone; a test file ends with `return Suite.run()`.

## Framework

`pass(name)` and `fail(name, detail)` are the primitives, `check` routes through them, and `eq` / `deep_eq` are now assertions rather than comparators — they record through pass/fail and a failure carries both values in the detail, which already made the test files shorter and the failures readable. `names` moved out of the shared framework onto the roles env, as it was role specific.

## Stubs

- `extend_requires`, `extend_script`, `extend_game`, `extend_defines`, all backed by a recursive merge which adds missing values and raises when one already exists, so nothing mutates the tables directly.
- Strict labels carry the factorio class names: `LuaBootstrap`, `LuaGameScript`, `LuaPlayer <name>`, `LuaPlayer <server>`.
- `script.register_metatable` records what was registered, and `stubs.save_load()` copies the script data the way factorio saves it: functions are refused and only registered metatables survive the copy. The roles env wraps it as `env.save_load()` which also calls `Roles.on_load()`. The new test — "on_load restores the state after a save and load" — assigns roles, round trips, and checks methods and held roles still answer. It genuinely depends on the module registering its metatable: removing the `script.register_metatable` call from control.lua fails it.

## Test naming

Both sides are named after what they cover, clusterio style. The lua tests read as `get_ordered_roles returns the most privileged first` and `role:assign applies locally and is sent to the controller`, so a scan of the names against the module is a coverage check. The javascript files wrap their tests in `class ControllerPlugin` / `class InstancePlugin` with subtests named per method (`handleSeedRolesRequest creates the roles and reuses them by name`), the message tests are one subtest per class, and `module.test.js` is renamed `control.test.js` after the file it covers. The dev cluster runs stay process based, as you said.

1159 assertions passing, lint 0 findings.

exp_roles/test/lua/env.lua
```diff
@@ -0,0 +1,91 @@
...
+    env.Roles = assert(loadfile(plugin_root .. "/module/control.lua"))()
+    env.Roles.on_server_startup()
+
+    --- Get a role by name, failing the test when it does not exist
+    function env.R(name)
+        return (assert(env.Roles.get_role_by_name(name), "No role named " .. name))
+    end
+
+    --- Names of an array of roles
+    function env.names(roles)
```
Cooldude2606: From looking closer at where this is used, names can be made more generic and included in the suite. But it should use more generic variable names to make it clear that it works for any table with a name property. Such as players, forces, surfaces, events, roles, commands, etc.

test/lua/stubs.lua
```diff
@@ -0,0 +1,177 @@
...
+        events = Stubs.strict("defines.events", {
+            on_player_joined_game = 1,
+            on_multiplayer_init = 2,
+        }),
+    })
+
+    local players_by_name, players_by_index, connected = {}, {}, {}
+    game = Stubs.strict("LuaGameScript", {
+        tick = 1,
+        players = players_by_name,
```
Cooldude2606: To work the same way as factorio we need to have a meta table on this that allows indexing by player id.

test/lua/stubs.lua
```diff
@@ -0,0 +1,177 @@
...
+
+    local next_event_id = 100
+    local registered_metatables = {} --- @type table<table, true>
+    script = Stubs.strict("LuaBootstrap", {
+        generate_event_name = function()
+            next_event_id = next_event_id + 1
+            return next_event_id
+        end,
+        raise_event = function(id, data)
+            stubs.events[#stubs.events + 1] = { id = id, data = data }
```
Cooldude2606: We dont need to store id seperatly, rather data.name should be set to the id, as well as data.tick to the current game tick.

exp_roles/test/lua/env.lua
```diff
@@ -0,0 +1,91 @@
...
+        env.Roles.initialise{
+            permission_names = permission_names,
+            roles = role_records(),
+            assignments = assignments or {},
+        }
+    end
+
+    --- Save and load the map, as far as the roles module can tell
+    local stub_save_load = env.save_load
+    function env.save_load()
```
Cooldude2606: We dont need this as the test should manually call on_load as this is inline with how we manually call other event handlers.

exp_roles/test/lua/assignment.lua
```diff
@@ -0,0 +1,125 @@
...
+    env.Roles.set_emit_events(true)
+    env.reset_log()
+    return players
+end
+
+test("role:assign applies locally and is sent to the controller", function(env)
+    local players = setup(env)
+    env.R("Moderator"):assign(players.bob, { by_player_name = "alice" })
+    check(#env.sent == 1 and env.sent[1].channel == "exp_roles:assignment_update", "the assignment is sent")
+    check(env.sent[1].data.name == "bob", "the payload names the player")
```
Cooldude2606: We should be using deep_eq for sent messages and factorio events.

exp_roles/test/controller.test.js
```diff
@@ -0,0 +1,198 @@
...
+
+	const roleStore = new lib.SubscribableDatastore(
+		...await new lib.MemoryDatastoreProvider(new Map(roles.map(role => [role.id, role]))).bootstrap()
+	);
+	const userStore = new Map(users.map(user => [user.name, user]));
+
+	const controller = {
+		config: { get: key => configValues[key] },
+		roles: roleStore,
+		users: {
```
Cooldude2606: Why do we use fake users and fake user manager when we import the real versions from controller. The controller itself can also be used directly without side effects if start is not called.

Cooldude2606 (changes requested):

bbassie:
Pushed e6552a39 covering all six points.

- **Real controller.** The tests construct a real `Controller` (unstarted, so side effect free, matching how the core tests do it) with real users through its `UserManager`. The only fakes left are two spies which record broadcasts and permission pushes on their way through to the real behaviour. This surfaced something the fakes had hidden: `getByNameMutable` and `valuesMutable` return a fresh `User` wrapper each call, so held references go stale after a write — the tests re-fetch, and set online time through `instanceStats` the way it actually arrives. It also meant registering the plugin's messages with `Link`, since the real controller validates them. One knock-on: any user update applies auto assignment immediately, which made the "granted on leave" assertion vacuous — `onPlayerEvent` is now observed directly instead.
- **Events.** `raise_event` fills in `name` and `tick` on the payload the way factorio does, rather than wrapping it, so a recorded event is exactly what a handler would receive.
- **Players by index.** `game.players` finds players by index as well as name through a metatable.
- **`names` generalised** onto the suite — it works for anything with a `name` property — and the env is back to fixtures only.
- **`save_load` no longer calls `on_load`**; the test calls the handler itself, same as the other handlers.
- **`deep_eq` for sent messages and events**, compared whole. That pins the full event shape including the tick, and that a rollback's payload has no `unassign` key rather than an empty one.

1151 assertions passing, lint 0 findings.

exp_roles/test/lua/env.lua
```diff
@@ -0,0 +1,75 @@
...
+            permission_names = permission_names,
+            roles = role_records(),
+            assignments = assignments or {},
+        }
+    end
+
+    return env
+end
+
+return Framework.new_suite(make_env)
```
Cooldude2606: I think we should make this:
```lua
return Framework.suite(function(env)
    env.Roles = --[[ omited ]]

    return env
end)
```

exp_roles/test/module/env.lua
```diff
@@ -0,0 +1,75 @@
...
+--- Create an independent environment with a fresh copy of the roles module
+local function make_env()
+    local env = Stubs.new()
+
+    env.Roles = assert(loadfile(plugin_root .. "/module/control.lua"))()
+    env.Roles.on_server_startup()
+
+    --- Get a role by name, failing the test when it does not exist
+    function env.R(name)
+        return (assert(env.Roles.get_role_by_name(name), "No role named " .. name))
```
Cooldude2606, line 71: This can be quite costly due to it doing a full search; since we contol the id value above, we can have a lookup index to directly convert from name to id then use get_role.

exp_roles/test/lua/assignment.lua
```diff
@@ -0,0 +1,149 @@
...
+local test, check, eq, deep_eq = Suite.test, Suite.check, Suite.eq, Suite.deep_eq
+
+--- alice is a moderator and bob a regular, with changes sent to the controller
+local function setup(env)
+    local players = {
+        alice = env.add_player("alice", 1),
+        bob = env.add_player("bob", 2),
+    }
+    env.initialise{
+        env.assignment("alice", { 5 }),
```
Cooldude2606: I dont like that we are using ids here, please instead get the role then access its id.

test/lua/stubs.lua
```diff
@@ -0,0 +1,184 @@
...
+    game = Stubs.strict("LuaGameScript", {
+        tick = 1,
+        players = players,
+        connected_players = connected,
+        print = function(message) stubs.printed[#stubs.printed + 1] = message end,
+        get_player = function(key) return players[key] end,
+    }, { "player" })
+
+    --- Add a player to the stubbed game
+    function stubs.add_player(name, index, is_connected)
```
Cooldude2606: Index is always continuous, so I think it is better if we calculate the next index rather than take it as a param.

test/lua/framework.lua
```diff
@@ -0,0 +1,153 @@
...
+    function Suite.check(ok, name, detail)
+        if ok then
+            Suite.pass(name)
+        else
+            Suite.fail(name, detail)
+        end
+    end
+
+    --- Check two values or arrays are shallowly equal
+    function Suite.eq(a, b, name)
```
Cooldude2606: I am not sure of any time were we would require shallow equal. So let's remove shallow eq, and rename deep eq to just eq.

exp_roles/test/instance.test.js
```diff
@@ -0,0 +1,146 @@
...
+
+const sampleRoles = () => [
+	new messages.RoleRecord(1, "Player", ["exp_scenario.gui.readme"], new messages.RoleMetaRecord(1, 2), true),
+	new messages.RoleRecord(5, "Moderator", ["exp_scenario.gui.readme", "core.admin"], new messages.RoleMetaRecord(5, 1)),
+];
+const sampleAssignments = () => [new messages.AssignmentRecord("alice", new Set([5]))];
+
+/** Build a plugin around a fake instance, started on first use by each test. */
+async function startPlugin(t2, { syncMode = "bidirectional", roles = sampleRoles() } = {}) {
+	const state = { sent: [], rcons: [], warnings: [], failNext: null };
```
Cooldude2606, line 49: Should use an real instance class here with spy methods just lile the contorller plugin tests.

Cooldude2606 (changes requested):
For function names in test names please format as `nane()` and for class methods use `.name()` for example `.get_player_names()` params should not be included. Following this review I am expecting this will be ready to merge, so please check you have covered as many code branches as are reasonable within the framework we have established.

bbassie:
Pushed 3218351d covering all six points, plus a coverage sweep since you expect this to be the last round.

- **Real instance.** The instance tests construct a real `Instance` with a real `InstanceConfig` (the plugin's config fields registered through `addPluginConfigFields`, so `exp_roles.sync_mode` is a real field), mirroring the controller tests. The spies are `sendTo` and the server: everything else is the real class, so `sendRcon` passes through the real script command gate on its way to the recorded server.
- **`eq` is deep.** Shallow eq is gone, `deep_eq` is now `eq`, all call sites updated.
- **`Framework.suite(extend_env)`** in the shape you sketched: the framework creates the stubs itself and hands each test the extension of a fresh set, so the plugin's env.lua is only the extension plus fixtures.
- **`add_player` assigns continuous indexes** itself; call sites pass only the name and connected state.
- **No raw ids in tests.** Assignments take role names, ids in expectations come from the role object (`env.R("Moderator").id`), and `env.R` resolves fixture names through an index rather than searching — falling back to the search only for roles a test created itself.
- **Naming**: functions are `name()` and methods `.name()` throughout, both lua and ts.

For the coverage sweep I walked the branches of control.lua and the two plugin classes against the test names; the gaps that were reasonable to close in this framework are now covered: sort tie breaking on the id, `get_player_highest_role()` erroring when a player has no roles, `player_has_all_permission()` for the server, `.assign()` defaulting the acting player from `game.player`, deleted assignment records being skipped on `initialise()`, `set_emit_events()` with no argument, `.print()` to a role with no online holders, `.handleRoleMetaUpdateRequest()` validation, `.handleAssignmentSubscription()` replay, unassignment over the ipc, and a refused removal having nothing to roll back. The branches I judged not reasonable to reach here: the `reject_assignment` path for a player object which has since left the game, and the seed path when no default role is configured — both need states the public api will not produce.

1163 assertions passing, lint 0 findings.

exp_roles/test/lua/assignment.lua
```diff
@@ -0,0 +1,161 @@
...
+        {
+            name = env.Roles.events.on_player_roles_changed,
+            tick = 1,
+            player_index = players.bob.index,
+            by_player_index = players.alice.index,
+            assigned = { moderator.id },
+            unassigned = {},
+        },
+    }, "the event carries the change")
+    check(#env.printed == 1 and env.printed[1][1] == "exp-roles.game-message-assign", "the change is announced")
```
Cooldude2606: This should be an eq.

exp_roles/test/lua/assignment.lua
```diff
@@ -0,0 +1,161 @@
...
+    eq(env.sounds, { "bob:utility/achievement_unlocked" }, "the assign sound plays")
+
+    local sd = env.Roles._script_data()
+    eq(sd.local_players.bob, { moderator.id }, "the role is held locally")
+    eq(sd.pending.bob, { moderator.id }, "the role is pending")
+
+    env.reset_log()
+    moderator:assign(players.bob)
+    check(#env.sent == 0 and #env.events == 0, "assigning a held role is a no-op")
```
Cooldude2606: When checking length is zero, eq with an empty table is better then length check so the failure details contain the actual data. I expect this will be common a pattern but also isnt obvious why an empty table is used. Therefore I propose we add `empty(value, name)` which uses `next(value) == nil` and `repr`

exp_roles/test/lua/assignment.lua
```diff
@@ -0,0 +1,161 @@
...
+end)
+
+test(".assign() defaults the by player to game.player", function(env)
+    local players = setup(env)
+    game.player = players.alice
+    env.R("Jail"):assign(players.bob, { silent = true })
+    check(env.events[1].by_player_index == players.alice.index, "the acting player is game.player")
```
Cooldude2606: This should be eq; this is the second occurrence so I will not comment on additional ones. Using eq rather than == provides more detailed failure messages.

exp_roles/test/lua/holders.lua
```diff
@@ -0,0 +1,68 @@
...
+    eq(Suite.sorted(Suite.names(moderator:get_players())), { "alice", "dave" }, "players are limited to this map")
+    eq(Suite.names(moderator:get_players(true)), { "alice" }, "filtered to connected players")
+    eq(Suite.names(moderator:get_players(false)), { "dave" }, "filtered to offline players")
+end)
+
+test(".print() reaches online holders", function(env)
+    setup(env)
+    env.reset_log()
+    check(env.R("Moderator"):print("hello") == 1, "print returns the number of players reached")
+    check(#env.printed == 1 and env.printed[1].to == "alice", "only online holders are reached")
```
Cooldude2606: Should use eq so we can confirm the message content.

exp_roles/test/lua/lookup.lua
```diff
@@ -0,0 +1,69 @@
...
+
+test("get_role_by_name() searches the roles", function(env)
+    env.initialise{}
+    check(env.Roles.get_role_by_name("Moderator") == env.R("Moderator"), "the role is found by name")
+    check(env.Roles.get_role_by_name("Gone") == nil, "deleted roles are not known")
+end)
+
+test("get_roles() returns every role", function(env)
+    env.initialise{}
+    check(#env.Roles.get_roles() == 5, "every role is returned with deleted roles skipped")
```
Cooldude2606: Should use Suite.sorted and Suite.names here

exp_roles/test/lua/players.lua
```diff
@@ -0,0 +1,93 @@
...
+    check(env.Roles.get_player_highest_role(players.alice) == env.R("Moderator"), "the highest held role")
+    check(env.Roles.get_player_highest_role(players.carol) == env.R("Player"), "the default role when none are held")
+end)
+
+test("get_player_highest_role() errors when a player has no roles", function(env)
+    local players = setup(env)
+    env.Roles.receive_role_updates{
+        { id = env.R("Player").id, name = "Player", permissions = {}, meta = { id = 0, order = 5 }, is_deleted = true },
+    }
+    check(not pcall(env.Roles.get_player_highest_role, players.carol), "no default role and no held roles errors")
```
Cooldude2606: This will be a false positive if the function fails for another reason, therefore we should test the returned error message. I expect this to be a common pattern so lets add `throws(fn, msg, name)`

Cooldude2606 (approved):
Just some final nits, once fixed happy to merge once I perform some local smoke tests.

bbassie:
Pushed 574a0a4b covering the nits.

- `Suite.throws(fn, message, name)` added and used for the highest role error, which now checks the message rather than just that something threw. `Suite.empty(value, name)` added, using `next` and reporting what the table held.
- Swept the rest: zero length checks are `empty`, scalar comparisons are `eq`, and the announced messages and print output are compared by content — the assign announcement, the server announcement, and `.print()` now pin the full message. `get_roles()` is asserted through sorted names rather than a count.

1171 assertions passing, lint 0 findings. Ready for your smoke tests.

Cooldude2606:
I have made some manual changes following my smoke test. Some changes were a result of the migration to emmy lua rather than because of this PR. As a final ask, could you please look at extending the type checking abilty of the test files so that autocomplete is supported. For example knowing `env.add_player` returns the mock player or that `env.R` returns a Role. I had a start at this and was able to get the stubs to auto compelte, but not the extensions made within the suite.

bbassie:
Pushed 08fe6b68 on top of your changes.

The missing piece for the extensions was making the suite generic over its environment: `Framework.suite` is `@generic T : Stubs` with the extension returning `T`, and `Suite<T>` forwards `T` into every test function. A test file declares `local Suite = ... --- @type Suite<ExpRoles.TestEnv>` and `env` autocompletes from there — emmylua instantiates the generic, so `env.add_player` returns `Stubs.Player`, `env.R` returns `ExpRoles.Role`, and `env.Roles` is the module. The stubs also declare their recorded fields (`events`, `printed`, `sent`, `sounds`) with their shapes.

Worth mentioning: the typing paid for itself the moment it landed, flagging real drift in the tests which I have fixed — `_script_data` is package private so the env now exposes `env.script_data()` instead, the join event handler was being called with a partial event (now carries its name and tick), the assignment record type had no `is_deleted`, `admin_state` was an undeclared field on the env (now a file local), and the event assertions which indexed `[1]` without a nil guard now compare the whole event list, which also pins the payloads more tightly. I verified the resolution end to end with a scratch file: a bogus field on the env, on a stub player, and on a role are each flagged as undefined-field.

Your smoke test changes all look good from my side. 1079 assertions passing, lint 0 findings.

Cooldude2606 (approved):

PR #455, opened 2026-09-05
Localization

### LINKED ISSUE #424: locale suggestions

PHIDIAS0303:
Localization changes.
Readme added 6.5.0, translate and wiki.
#424


Cooldude2606 (changes requested):
Please do not merge the legacy cfg files. They are kept seperate so it is clear what still needs to be migrated.

Also I like that you have removed the repetitive "enter" and "exit" strings; however, "tooltip" is a bad name for it because these are commands and commands do not have tooltips. I would suggest "selection-name"

Within the readme can the release be named "Clusterio Plugin (Beta)" as I do intended 7.0.0 to be the main release of the plugin. I might even look at unpublishing it and making it an actual pre-release.

PHIDIAS0303:
there is a comment to seperate the files still.

Cooldude2606:
> there is a comment to seperate the files still.

Does you tool require them to be in a single file?

PHIDIAS0303:
yes because it can only mark one base file

Cooldude2606 (approved):

PR #456, opened 2026-09-05
Remove module inserter

PHIDIAS0303:
can be replaced by base function


Cooldude2606:
Not sure this needs to be removed as it is still a good user interface and is quicker to use than the regular upgrade planner. Do you have another reason for why it should be removed?

PHIDIAS0303:
it is not dynamic slot.
it does not include non quality space age right now.
it will be better for using native one

PR #457, opened 2026-09-05
Move the seed into exp_scenario and seed the permission groups

bbassie:
Follow up to #453. The seed roles were owned by exp_roles but describe the scenario, so they now live in `exp_scenario/seed.ts` next to the permissions they grant. Each seed role names the permission group its holders belong to, and the five groups the legacy config defined (Admin, Trusted, Standard, Guest, Restricted) are seeded through exp_groups. Roles which follow each other in rank with the same group share a mapping, so the seed makes one mapping per group. Mapping priorities follow how exp_roles ranks a player's highest role (priority, then order), so Jail lands in Restricted no matter what else the player holds. System has no group and stays in Factorio's Default group, matching the old `set_permission_group("Default", true)`.

`SeedRolesRequest` becomes `SeedRequest` in exp_scenario behind a new `exp_scenario.seed` permission, and the button moves to the scenario's web plugin (which gains a web entrypoint), still rendered on the roles page. Re-seeding reuses roles by name and only adds permissions, resets groups by name to the seed, reuses a mapping holding exactly the same roles, and skips priorities held by other mappings.

With groups seeded from the controller the legacy `expcore.permission_groups` module, its config, the `_file_loader` entry and the `Group` rcon static are removed. exp_roles no longer needs its devDependency on exp_scenario; exp_scenario depends on `@expcluster/roles` and `@expcluster/permission-groups` instead.

Tests moved with the seed and grew a controller test that runs the seed against a real `Controller` with both sibling plugins loaded (`pnpm --filter @expcluster/scenario test`, 592 assertions; exp_roles still passes with 523). Lua lint is clean.

Smoke tested on my dev cluster against a database that already had the roles and a hand made group mapped at priority 100: the seed created 5 groups and 14 mappings (now 5, one per group, after the review), left the priority 100 mapping alone, re-seeding was idempotent, and the instance showed all five groups in game with the expected allowed action counts (Restricted allows only `write_to_console`).


exp_scenario/controller.ts
```diff
@@ -1,6 +1,157 @@
...
+		for (const seedRole of ranked) {
+			priority += 1;
+			while (taken.has(priority)) {
+				priority += 1;
+			}
+
+			const roleId = roleIds.get(seedRole.name)!;
+			mappings.push(new RoleMappingRecord(
+				existing.get(roleId)?.id ?? newId(groupsPlugin.roleMappings),
+				new Set([roleId]),
```
Cooldude2606: Mappings have the abilty to contain multiple roles. We should make use of this to have have roles which share a priority and group to also share a mapping. For example `Senior Administrator` `Administrator` `Moderator` and `Trainee` are all in the `Admin` group and same priority, so only a single mapping is required.

bbassie: Done in d342dee7. Roles which follow each other in rank with the same group now share a mapping, so the seed makes one per group (Admin, Trusted, Standard, Guest, Restricted). Re-seeding reuses a mapping that holds exactly the same roles.

Cooldude2606:
Apologies, you were right the first time as `A mapping applies only if the user has all listed roles.`
You can revert the last change and then I'll merge.

exp_scenario/controller.ts
bbassie: Reverted, the branch is back to one mapping per role. I checked `computeResolvedAssignment` and it requires the player to hold every role in `mapping.roleIds`, so a shared Admin mapping would only have matched someone holding all four at once.

Cooldude2606 (approved):

PR #458, opened 2026-09-05
Move jail into exp_scenario

bbassie:
With the role system finished the jail module had nothing legacy left in it, so it moves to `exp_scenario/module/control/jail.lua` and the copy in exp_legacy is removed. It has no event handlers of its own, so it is required by its callers rather than added in `control.lua`.

Small api tidy while moving: it takes `LuaPlayer` only (the player list passed names before), `reason` is required since every caller already supplies one, `jail_player`/`unjail_player` return `false` rather than `nil` when there is nothing to do, and the event ids are exposed as `Jail.on_player_jailed` and `Jail.on_player_unjailed` in the same shape as exp_util's selection events. The five callers (jail command, protection jail, report jail, discord alerts, player list actions) are updated.

Lua lint is clean. Loaded on my dev instance together with the other scenario moves: 0 load errors and `/jail` and `/unjail` registered.


Cooldude2606 (approved):
Only needed to remove redudant event properties which factorio sets for you.

PR #459, opened 2026-09-05
Move protection into exp_scenario

bbassie:
The entity protection module moves to `exp_scenario/module/control/protection.lua` in the event_handler shape the other control modules use, and the copy in exp_legacy is removed. The config stays in `exp_legacy/config/protection.lua` for now, like nuke_protection.

Changes while moving: the config lookups are built as local sets instead of rewriting the config tables in place, the two `on_pre_player_mined_item` handlers become one, the entity and area key helpers are exposed so `protected_entities.lua` no longer keeps its own copies (that was a TODO), and the event ids are `Protection.on_player_mined_protected` and `Protection.on_repeat_violation`. The three callers (protection command, protection jail, discord alerts) are updated and the module is added in `control.lua`.

Lua lint is clean. Loaded on my dev instance together with the other scenario moves with 0 load errors.

This touches the line next to the one the jail PR changes in `protection_jail.lua`, so whichever merges second will need a one line rebase, which I will do.


Cooldude2606 (approved):
I have fixed the conflict, and also made the same change I did in Jail for the event name being redundant.

PR #461, opened 2026-09-05
Remove the warning system

bbassie:
Warnings were a moderator tool layered on top of reports and nothing used them beyond their own command and the player list button. This removes the legacy module and its config, the exp_scenario command and its five permissions, the seed entries, the discord alert block and config flag, the player list button, and the locale sections in en, zh-CN and zh-TW (scenario and legacy).

Reporting from the player list used to be hidden from players who could give warnings. Now anyone can report a player who is not report immune, which is what the command already allowed.

exp_roles tests pass (the seed test checks every seeded permission is still defined) and Lua lint is clean. Loaded on my dev instance together with the other scenario moves with 0 load errors and the warning commands no longer registered.

The seed edit lands in `exp_roles/seed.ts` here and moves to `exp_scenario/seed.ts` in the seed groups PR; git follows the rename on rebase. It also touches the line next to the jail PR's change in `player_list_actions.lua`, so whichever merges second needs a one line rebase, which I will do.


Cooldude2606 (approved):

Cooldude2606 (approved):
Fixed the conflict on this branch. Seed roles will need to rebase.

PR #464, opened 2026-09-05
Count kills and damage dealt from vehicles

### LINKED ISSUE #219: Kills done while in a vehicle arn't tracked
**Factorio version**
Version number: 1.1.37

**Scenario version**
Version number: 6.2.0

**Bug description**
The datastore statistics for player kills are not increased when the player is in a vehicle. The issue is because the filter during on_entity_died is too restrictive. There is a good chance that this issue also effects the on_entity_damaged event as this uses a similar filter.

**Reproduction method**
1. Get in any vehicle, such as a spidertron.
2. Kill biters.
3. Counter does not increase

**Expected behavior**
The counter should track all player kills.

bbassie:
Fixes #219.

The `DamageDealt` and `Kills` statistics only counted when `event.cause` was a character, so anything done from a car, tank, spidertron or locomotive was dropped. The cause is now resolved to a player through a `get_cause_player` helper: characters map to their player as before, and cars, spider vehicles and locomotives map to their driver, or their passenger when nobody is driving. Turrets, artillery and uncrewed trains still count for nobody, which matches the conclusion in the issue thread that automatic kills should not be credited.

Lua lint is clean and the scenario loads on my dev instance. I could not stage a vehicle kill headless, so the credit path itself is untested in game.

### DIFF
--- a/exp_legacy/module/modules/data/statistics.lua
+++ b/exp_legacy/module/modules/data/statistics.lua
@@ -113,16 +113,47 @@ if config.MachinesRemoved or config.TreesDestroyed or config.OreMined then
     Event.add(defines.events.on_player_mined_entity, on_event)
 end
 
+--- Vehicles which credit their driver, or failing that their passenger, with what they hit
+local crewed_vehicles = { car = true, ["spider-vehicle"] = true, locomotive = true }
+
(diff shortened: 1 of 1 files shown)


Cooldude2606 (approved):
Test locally including tank and running things over rather than using a weapon. All works as expected.

PR #465, opened 2026-09-05
Only log a rocket as fired when its ammo count drops

### LINKED ISSUE #242: False positives for rocket fired event log
**Factorio version**
Version number: 1.1.87

**Scenario version**
Version number: 6.2.1

**Bug description**
There are a few cases of false positives for the rocket/nuke fired event in the decon log. The cause is because there is no "fired" event in factorio, rather we rely on `on_player_ammo_inventory_changed` which can be raised by equiping rockets for the first time or by switching between rockets and explosive rockets.

**Reproduction method**
1. Have both rockets and explosive rockets in your inventory.
2. Equip the rockets, this will emit an event in the log.
3. Equip the explosive rockets in the same slot, this will raise another event in th log.
4. Keep switching between the two, a new log line is added each time.

**Expected behavior**
Log events should only be emited when a rocket is fired.

bbassie:
Fixes #242.

There is no fired event, and the deconstruction log wrote a line on every `on_player_ammo_inventory_changed`, which fires when rockets are loaded for the first time and when swapping between rocket types. The log now keeps the last seen name and count of each of the player's ammo slots (in storage, per player, cleared when they leave) and only records a shot when a slot holds the same ammo with one fewer, or empties from a single round. Loading, swapping and picking ammo up all change the name or increase the count, so none of them log.

Two smaller changes fell out of this: every ammo slot is checked rather than only the selected gun, and only the three configured ammo types (rocket, explosive rocket, atomic bomb) are logged, where before any ammo change produced a `shot-<ammo>` line.

Lua lint is clean and the scenario loads on my dev instance. Firing itself needs a player, so the count logic is untested in game.

### DIFF
--- a/exp_scenario/module/control/deconstruction_log.lua
+++ b/exp_scenario/module/control/deconstruction_log.lua
@@ -3,6 +3,7 @@ Log certain actions into a file when events are triggered
 ]]
 
 local ExpUtil = require("modules/exp_util")
+local Storage = require("modules/exp_util/storage")
 local Roles = require("modules/exp_roles")
 local config = require("modules.exp_legacy.config.deconlog")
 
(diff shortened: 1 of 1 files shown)


exp_scenario/module/control/deconstruction_log.lua
```diff
@@ -130,29 +131,63 @@ local function on_player_mined_entity(event)
...
+            add_log_line(player, "shot-" .. fired, format_position(player.physical_position), format_position(player.shooting_state.position))
+        end
     end
+end
 
-    add_log_line(player, action_name, format_position(player.physical_position), format_position(player.shooting_state.position))
+--- Forget the ammo of a player who left, their slots are read again on the next change
+--- @param event EventData.on_player_left_game
+local function on_player_left_game(event)
+    ammo_slots[event.player_index] = nil
```
Cooldude2606, line 205: Because we clear on leave, this means that the first shot is never logged. This is a gap in coverage. We need to repopulate the table on player join before their ammo count changes for the first time.

Cooldude2606 (changes requested):

Cooldude2606 (approved):

PR #468, opened 2026-09-05
Move the community links into a config file

### LINKED ISSUE #424: locale suggestions

bbassie:
Fixes #424.

The five community links were locale strings repeated in every language. They are now `exp_legacy/config/links.lua`, read by the readme's external links table and the join greeting, and the `[links]` sections are removed from all three locales.

For the selection mode messages I kept per command keys, since each carries its own hint, but fixed the wording where it was wrong or empty: the lawnmower said it would fill the area with water, repair and blueprint clearing had a generic "select the area", and artillery had "Existed" for exited. If you would rather have one shared section in exp_util's selection module with the command name as a parameter, say so and I will follow up.

Lua lint is clean, and on my dev instance the config loads and the readme resolves the links from it.

### DIFF
--- a/exp_legacy/module/config/links.lua
+++ b/exp_legacy/module/config/links.lua
@@ -1,0 +1,10 @@
+--- Links to the community, shown in the readme and the join message
+-- @config Links
+
+return {
+    discord = "https://discord.explosivegaming.nl",
+    website = "https://www.explosivegaming.nl",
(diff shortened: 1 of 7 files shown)


Cooldude2606:
This overlaps with #455 so only one of these two will be merged.

PR #470, opened 2026-09-06
fix readme

PHIDIAS0303:


Cooldude2606 (changes requested):
It doesnt look very good with single word links. Can you add a sentence for each.

Cooldude2606 (approved):

PR #473, opened 2026-09-09
Reference exp_roles and exp_groups from the exp_scenario build

bbassie:
Since #457 exp_scenario imports types from `@expcluster/roles` and `@expcluster/permission-groups`. Their declaration files only exist once the two packages have been built, so building exp_scenario before them fails with `TS2307: Cannot find module '@expcluster/roles/dist/node/controller'` and four more like it. A plain `pnpm install` gets away with it because pnpm runs the prepare scripts in dependency order, but a filtered install, or `tsc --build` inside exp_scenario on a fresh clone, hits the error.

Two commits:

1. Project references from exp_scenario's node tsconfig to the node tsconfigs of exp_roles and exp_groups, the same way the core clusterio plugins reference `packages/lib`. `tsc --build` now builds those two first when they are out of date.
2. exp_roles and exp_groups re-export their messages and the `ControllerPlugin` type from the package index, so exp_scenario imports `from "@expcluster/roles"` instead of `from "@expcluster/roles/dist/node/messages"`. The controller export is type only, so nothing from `@clusterio/controller` ends up in the web bundle (checked the emitted index chunk). The test keeps the deep require for the controller classes since those cannot be re-exported at runtime.

Verified by deleting every `dist` folder and running `tsc --build` in exp_scenario alone, then at the repo root, plus webpack for all three packages and the roles and scenario tap suites.

### DIFF
--- a/exp_groups/index.ts
+++ b/exp_groups/index.ts
@@ -1,6 +1,9 @@
 import * as lib from "@clusterio/lib";
 import * as messages from "./messages";
 
+export * from "./messages";
+export type { ControllerPlugin } from "./controller";
+
 declare module "@clusterio/lib" {
 	export interface InstanceConfigFields {
 		"exp_groups.sync_mode": "enabled" | "disabled" | "bidirectional"
--- a/exp_roles/index.ts
+++ b/exp_roles/index.ts
@@ -1,6 +1,9 @@
 import * as lib from "@clusterio/lib";
 import * as messages from "./messages";
 
+export * from "./messages";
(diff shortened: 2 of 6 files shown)


Cooldude2606 (changes requested):
We cannot export the plugins from the index file as this prevents it being loaded into other contexts which are not "standalone".

Cooldude2606 (approved):
Nvm I failed to see that it was type export only. LGTM.

